PR #131, opened 2023-03-20
gitbook dev for CLI

taranvohra:
https://github.com/GitbookIO/pca-squad/issues/68

This PR adds a `gitbook dev` command for the CLI. The command does the following in a nutshell:

- Build the script and serve it through miniflare dev server
- Use `cloudflared` to create a tunnel mapped to localhost
- Store the tunnel in GitBook platform so that we can receive integration events locally
- Use `chokidar` to watch for changes in integration script and hot reload the dev server

It also adds support for a `gitbook-dev.yaml` config file for starting the dev server for a particular space

Finally, since Cloudflare does not have a official NPM package, we're using a `postinstall` script in the CLI to fetch the binary from [Cloudflare GH releases](https://github.com/cloudflare/cloudflared/releases/) and store it in `dist` to be used later in our dev server


packages/cli/postinstall.js
SamyPesse, line 1: Have you seen https://github.com/JacobLinCool/node-cloudflared ?

We could also only install `cloudflared` the first time the user is running `gitbook dev` instead of in `postinstall`.

packages/cli/src/build.ts
```diff
@@ -0,0 +1,63 @@
...
+ */
+export async function buildScriptFromManifest(manifestSpecPath: string): Promise<BuildOutput> {
+    const manifest = await readIntegrationManifest(manifestSpecPath);
+    /**
+     * Resolve the input and output file paths relatve to the manifest file.
+     */
+    const inputFilePath = resolveFile(manifestSpecPath, manifest.script);
+    const outputFilePath = resolveFile(
+        manifestSpecPath,
+        path.join(GITBOOK_OUTPUT_DIR, 'script.js')
```
SamyPesse, line 29: Where is this directory stored?

It's unclear from the `path.join(GITBOOK_OUTPUT_DIR, 'script.js')` missing a root, it looks like it'll use `process.cwd()`.

Why not use a temporary directory instead of polluting the local directory?

packages/cli/src/build.ts
```diff
@@ -0,0 +1,63 @@
...
+    );
+
+    await esbuild.build({
+        platform: 'neutral',
+        entryPoints: [inputFilePath],
+        outfile: outputFilePath,
+        bundle: true,
+        minify: true,
```
SamyPesse, line 37: In development, we probably don't want this? or we want sourcemap?

packages/cli/src/cli.ts
```diff
@@ -52,6 +53,14 @@ program
         await promptNewIntegration(path.resolve(process.cwd(), dirPath));
     });
 
+program
+    .command('dev')
+    .description('run the integrations dev server')
+    .option('-p, --port <number>', 'port to run the dev server on')
```
SamyPesse, line 59: How can the user chooses the space to dev with?

SamyPesse, line 59: What port/server ? why should the end user know this?

It feels like I'll have to access this port, when it's not true. Why not use a random port?

packages/cli/src/dev.ts
```diff
@@ -0,0 +1,143 @@
...
+import ora from 'ora';
+import * as path from 'path';
+
+import { buildScriptFromManifest } from './build';
+import { prettyPath } from './files';
+import { getDefaultManifestPath, resolveFile, resolveIntegrationManifestPath } from './manifest';
+import { getAPIClient } from './remote';
+import { createDevTunnel } from './tunnel';
+
+export const GITBOOK_DEV_SERVER_PORT = 8787;
```
SamyPesse, line 16: This port might be taken, there is a node module to find an empty port. I think it'll be better than having a user option.

packages/cli/src/dev.ts
```diff
@@ -0,0 +1,143 @@
...
+ * Start the integrations dev server on the given port. If no port is given,
+ * the default one is used. The dev server will automatically reload changes to
+ * the integration script.
+ */
+export async function startIntegrationsDevServer(port: number = GITBOOK_DEV_SERVER_PORT) {
+    spinner.start('Starting dev server...');
+    /**
+     * Read the integration manifest and build the integration script so that
+     * it can be served by the dev server.
+     * Also, read the dev config file to get the space to use for the dev server.
```
SamyPesse, line 31: Can it be specified by the cli arguments instead?

Having to write my space in the manifest feels slow. Maybe `gitbook dev <spaceId>` could write it for me.

SamyPesse: Also you don't seem to handle the case the space is not defined to show a proper error to the user.

packages/cli/src/dev.ts
```diff
@@ -0,0 +1,143 @@
...
+
+    /**
+     * Start the miniflare dev server. It will automatically reload the script
+     * when it detects a change with watch mode
+     */
+    const mf = new Miniflare({
+        scriptPath,
+        port,
+        liveReload: true,
+        watch: true,
```
SamyPesse, line 52: You are watching the script generated by esbuild, but not the source.

You need your own watcher here, or you need to have miniflare triggers the build (like https://miniflare.dev/developing/esbuild)

packages/cli/src/dev.ts
```diff
@@ -0,0 +1,143 @@
...
+    /**
+     * Add the tunnel to the integration for the dev space events in the GitBook platform
+     */
+    const api = await getAPIClient(true);
+    const tunnel = await api.integrations.addIntegrationTunnelForSpace(manifest.name, {
+        space: devConfig.space,
+        url: tunnelUrl,
+    });
+
+    spinner.succeed(`Dev server started on ${port} 🔥`);
```
SamyPesse, line 65: We should add more explanation about what it means: "Every events related to the space ... will be dispatched to your locally running version of the integration".

packages/cli/src/dev.ts
SamyPesse: Missed it below!

packages/cli/src/dev.ts
```diff
@@ -0,0 +1,143 @@
...
+
+    /**
+     * Additionally, watch the directory of the script file for changes. When a change is
+     * detected, rebuild the script and miniflare will automatically reload it in the dev server.
+     */
+    chokidar
```
SamyPesse, line 71: Shouldn't we kill chokidar when `process.on('SIGINT')`?

packages/cli/src/dev.ts
```diff
@@ -0,0 +1,143 @@
...
+import * as path from 'path';
+
+import { buildScriptFromManifest } from './build';
+import { prettyPath } from './files';
+import { getDefaultManifestPath, resolveFile, resolveIntegrationManifestPath } from './manifest';
+import { getAPIClient } from './remote';
+import { createDevTunnel } from './tunnel';
+
+export const GITBOOK_DEV_SERVER_PORT = 8787;
+export const GITBOOK_DEV_CONFIG_FILE = 'gitbook-dev.yaml';
```
SamyPesse, line 17:
```suggestion
export const GITBOOK_DEV_CONFIG_FILE = '.gitbook-dev.yaml';
```

packages/cli/src/build.ts
taranvohra: The directory is stored in the same parent as the `gitbook-manifest.yaml` (look  `resolveFile` method's first argument)

As for the choice of directory `.gitbook`, again I am going with other popular CLIs like NextJS that uses a `.next` folder to output files during `next dev`

packages/cli/src/build.ts
taranvohra: Yes, I didn't give thought to this part but probably something along the lines of what you said

packages/cli/src/cli.ts
taranvohra: That's where `gitbook-dev.yaml` comes into picture

packages/cli/src/dev.ts
taranvohra: Agree with the suggestion 👍

packages/cli/src/dev.ts
taranvohra: I'm more inclined on not having it as a CLI option but use a `gitbook-dev.yaml` config to specify with easy to support it in the future too

taranvohra: > Also you don't seem to handle the case the space is not defined to show a proper error to the user.

The error would be thrown by the API when we are setting the tunnel if `space` does not exist. Would that not be enough? 😁

packages/cli/src/dev.ts
taranvohra: Sure 👍

packages/cli/src/dev.ts
taranvohra: Ah yes!

packages/cli/src/build.ts
SamyPesse: I don't think we can compare to Next, we are forcing the users to gitignore `.gitbook`.

Also `.gitbook` is a special folder when syncing with GitHub that should not be gitignored ;) so this PR will prevent the use of the GitSync.

packages/cli/src/dev.ts
SamyPesse: > The error would be thrown by the API when we are setting the tunnel if space does not exist. Would that not be enough? 😁

Yes, I was referring to the fact that the space isn ot even defined (empty string, undefined).

> I'm more inclined on not having it as a CLI option but use a gitbook-dev.yaml config to specify with easy to support it in the future too

I disagree, I think it adds friction, and makes it harder to test on multiple spaces (why would we want the user to use only one space for an integration?).

packages/cli/src/build.ts
taranvohra: Oh right, good point. How about `.gitbook-dev` folder? We can generate a `.gitignore` for this folder during `gitbook create` that adds `.gitbook-dev` to gitignore

packages/cli/src/dev.ts
taranvohra: > Yes, I was referring to the fact that the space isn ot even defined (empty string, undefined).
Oh hm, I thought it would be taken care of by `Ajv` when we validate the dev config yaml but I'll add another check in code

> I disagree, I think it adds friction, and makes it harder to test on multiple spaces (why would we want the user to use only one space for an integration?).
You can still test on multiple spaces. All it needs is a change in the dev config `space: newSpaceId`. 

Maybe I misunderstood your first question. We certainly can generate an empty `.gitbook-dev.yaml` during `gitbook create` but I'm not sure about polluting the command with CLI args because it can easily blow if this dev config was to expand

PR #132, opened 2023-03-29
Add organization and visibility to generated manifest

addisonschultz:
https://github.com/GitbookIO/pca-squad/issues/97

This PR adds `organization` and `visibility` to the generated `gitbook-manifest.yaml` file from the CLI. Organization is a required field in order to publish, and we want to set the visibility by default to private when developing locally. 

Additionally, it is not allowed for non-GitBookers to publish integrations publicly

### DIFF
--- a/packages/cli/src/init.ts
+++ b/packages/cli/src/init.ts
@@ -3,7 +3,7 @@ import * as fs from 'fs';
 import * as path from 'path';
 import prompts from 'prompts';
 
-import { IntegrationScope } from '@gitbook/api';
+import { IntegrationScope, IntegrationVisibility } from '@gitbook/api';
 
 import packageJSON from '../package.json';
 import { fileExists } from './files';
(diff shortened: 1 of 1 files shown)


packages/cli/src/init.ts
```diff
@@ -34,6 +34,12 @@ export async function promptNewIntegration(dirPath: string): Promise<void> {
             message: 'Title of the integration:',
             initial: (prev) => prev || '',
         },
+        {
+            type: 'text',
+            name: 'organization',
+            message: 'ID of the organization this integration is owned by',
```
taranvohra, line 40: It can also be the GitBook `subdomain`, probably worth mentioning it

taranvohra (approved):

packages/cli/src/init.ts
addisonschultz: Nice, didn't know you already updated that. I'll update and merge

addisonschultz:
@taranvohra This needs to be published to `npm` though, right? I've never done a publish - but think there's more updates (like the `dev` command) that need to be published too, so this can be batched with that I suppose. 

Are you able to deploy that today?

taranvohra:
@addisonschultz Yes, the idea was to batch these updates in a single release. We can do one today

PR #133, opened 2023-03-30
Add description to CLI manifest generation

addisonschultz:
This PR adds `description` to the generated manifest. This is a required field to publish an integration. 

**Not** adding this to the CLI prompts, as I think it's better for users to go in after and write a more detailed description, and isn't really necessary to have a descriptive field during initial publish.


taranvohra (approved):
You should sync this branch, not sure why the other PR's updates are also coming here 😅

PR #168, opened 2023-05-29
Increase usefulness of Slack integration

tyronedougherty:
Hi GitBook team 👋 

I love having Slack notifications when spaces get updated, but at the moment there's not a lot of context for the update:

> **GitBook** [App]
> Content of **Example Space Name** has been updated

I think we can do better. I've updated the Slack integration to include the message for the update (if it exists) and the titles of all updated pages.

> **GitBook** [App]
> Content of **Example Space Name** has been updated: **Add background info on xyz**
>
> Pages updated:
> * API - xyz
> * Concept - xyz

Unfortunately there wasn't really a way for me to test whether this works or not, so I've made the changes entirely based on what the types tell me should exist and the Slack API spec for section blocks. If you decide to merge this, it should probably be tested first 😆 

There may also be a better way of laying this content out, but I'm lazy and figured this was the fastest approach.

Let me know what you think! Thank you

### DIFF
--- a/integrations/slack/src/index.ts
+++ b/integrations/slack/src/index.ts
@@ -27,6 +27,10 @@ const handleSpaceContentUpdated: EventCallback<
     }
 
     const { data: space } = await api.spaces.getSpaceById(event.spaceId);
+    const { data: revision } = await api.revisions.getRevisionById(event.spaceId, event.revisionId);
+
+    const revisionMessage = revision.git.message;
+    const pagesUpdatedInRevision = revision.pages.map((page) => page.title);
 
     await slackAPI(context, {
         method: 'POST',
@@ -40,7 +44,19 @@ const handleSpaceContentUpdated: EventCallback<
                         type: 'mrkdwn',
                         text: `Content of *<${space.urls.app}|${
                             space.title || 'Space'
-                        }>* has been updated`,
+                        }>* has been updated${
+                            revisionMessage.length > 0 ? `: *${revisionMessage}*` : ''
+                        }
+
+                        ${
+                            pagesUpdatedInRevision.length > 0
+                                ? `\n\nPages updated: \n ${pagesUpdatedInRevision.map(
+                                      (title) => `• ${title}\n`
+                                  )}`
+                                : ''
+                        }
+
+                        `,
                     },
                 },
             ],
--- a/.changeset/silly-glasses-jump.md
+++ b/.changeset/silly-glasses-jump.md
@@ -1,0 +1,5 @@
+---
+'@gitbook/integration-slack': minor
+---
+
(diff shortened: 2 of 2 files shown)


integrations/slack/src/index.ts
```diff
@@ -27,6 +27,10 @@ const handleSpaceContentUpdated: EventCallback<
     }
 
     const { data: space } = await api.spaces.getSpaceById(event.spaceId);
+    const { data: revision } = await api.revisions.getRevisionById(event.spaceId, event.revisionId);
+
+    const revisionMessage = revision.git.message;
+    const pagesUpdatedInRevision = revision.pages.map((page) => page.title);
```
taranvohra, line 33: Hello 👋, thanks for the contribution. It's really a neat idea. Though the solution will not work as expected because a `revision` does not contain only "updated pages" but instead it has reference to `all` the pages in the space. 

So, this is basically going to print `titles` of all the pages in the space which isn't what we want at the end of the day ;) 

On a related note, you can test integrations locally without having to publish again & again. Refer https://developer.gitbook.com/getting-started/development#local-development

tyronedougherty: Thanks for linking me to the docs!

Is there a way to get only the pages that have been updated for that revision? If not, the Git message is probably good enough to bring a bit more context 😄

taranvohra: Unfortunately no, not from the revisions. The Git message is also optional and present for only those spaces that has Git Sync set up 😄

tyronedougherty:
Damn.

Ok I misunderstood the purpose of that field as well then. Is there a way
to get the description of the change that people enter for each change
request?

Thanks for your help ☺️

taranvohra:
> Damn. Ok I misunderstood the purpose of that field as well then. Is there a way to get the description of the change that people enter for each change request? 

Not as of now

tyronedougherty:
> > Damn. Ok I misunderstood the purpose of that field as well then. Is there a way to get the description of the change that people enter for each change request?
> 
> Not as of now

Darn. Is there a place I can raise a feature request for these pieces of information? 

Having access to it would help increase the usefulness of the Slack integration massively. Right now nobody really understands what has changed without logging in and doing a diff. For larger teams it would be better to see a high level summary in Slack when the docs are updated.

addisonschultz:
@tyronedougherty You're more than welcome to raise the feature request in this repository as an issue! 

https://github.com/GitbookIO/integrations/issues

We are keeping track and organizing incoming requests and bugs here. You can use the labels to help organize your idea/request as well!

tyronedougherty:
Thank you @addisonschultz! I've raised feature request #177 for this issue. Though I don't seem to have permissions to update the labels to help organise.

I'll close this PR now 👍

PR #195, opened 2023-07-11
PostHog integration

thebiglabasky:


integrations/posthog/package.json
```diff
@@ -0,0 +1,18 @@
+{
+    "name": "@gitbook/integration-hotjar",
```
taranvohra, line 2: 😄 
```suggestion
    "name": "@gitbook/integration-posthog",
```

integrations/posthog/gitbook-manifest.yaml
```diff
@@ -0,0 +1,50 @@
...
+                type: string
+                title: Project API Key
+                description: Available in the PostHog installation snippet, as the first parameter, starting with 'phc_'
+            instanceAddress:
+                type: string
+                title: Instance address
+                description: Depends on your PostHog installation's location (either EU, or US)
+                enum:
+                    - https://eu.posthog.com
+                    - https://us.posthog.com
```
taranvohra, line 47: I feel using `US` and `EU` as enum values is a better user experience as looking at an url can be abit meh in that select dropdown.

So basically you capture `instanceLocation` (US/EU/etc) instead of `instanceAddress` and in your integrations code you obtain the instance address through a mapping

```js
const instanceAddresses = {
  us: 'https://us.posthog.com',
  eu: 'https://eu.posthog.com',
}

const instanceLocation = environment.spaceInstallation.configuration.instanceLocation;
const instanceAddress = instanceAddresses[instanceLocation.toLowerCase()];
```

taranvohra (approved):

integrations/posthog/gitbook-manifest.yaml
thebiglabasky: Right that's a better look and feel indeed. Let's do that instead

PR #198, opened 2023-07-13
Minor source formatting in docs

yuchi:
In code snippets there were minor source formatting problems:
- misaligned indentations
- incosistent object method shorthand usage

This commit fixes these


addisonschultz (approved):
Thanks for your contribution! I think it should be easier to copy some of these snippets now. 

For reference, the `action` property of the `createComponent` call can either be a named function or added as a key—Either one is valid. I'll add the change regardless 😄 

Thanks!

yuchi:
Yeah both syntaxes were valid but were incosistent. Thanks for the approval and merge

PR #199, opened 2023-08-02
Update CLI script to create more accurate `package.json`

dstotijn:
- `@gitbook/runtime` is moved to `dependencies`
- `@gitbook/eslint-config` is added to `devDependencies`
- `@cloudflare/workers-types` is removed from `devDependencies`
- Remove the GitBook-specific `publish-integrations-staging` script, replace with `publish` script
- Use a library for typing of `package.json` schema, makes code a bit more readable

### DIFF
--- a/packages/cli/src/init.ts
+++ b/packages/cli/src/init.ts
@@ -1,3 +1,4 @@
+import { JSONSchemaForNPMPackageJsonFiles2 as PackageJSON } from '@schemastore/package';
 import { spawn } from 'child_process';
 import detent from 'dedent-js';
 import * as fs from 'fs';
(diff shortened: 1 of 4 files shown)


packages/cli/src/init.ts
```diff
@@ -136,32 +137,25 @@ export async function initializeProject(
...
+    const packageJsonObject: PackageJSON = {
+        name: projectName,
+        private: true,
+        scripts: {
+            lint: 'eslint --ext .js,.jsx,.ts,.tsx .',
+            typecheck: 'tsc --noEmit',
+            publish: 'gitbook publish .',
+        },
+        dependencies: {
+            '@gitbook/runtime': 'latest',
```
dstotijn, line 149: I'm now doubting again if this should be a production dependency or not; can it remain a dev dep because `@gitbook/runtime` gets bundled into a JS file when the integration it published and deployed on our cloud infra? 🤔

taranvohra: I'm confused, isn't that the main reason why it should be in `deps`? 😄  because it _should_ be bundled with the main code as it'll be essential for the proper execution. Same reason why `react` is usually in `deps` but something like `eslint` isn't

dstotijn: Yeah, that's what I figured as well, hence the diff in this PR. But I was doubting it later because I noticed that the Formspree integration for instance depends on `@gitbook/runtime` but has it listed as a dev dependency ([src](https://github.com/GitbookIO/integrations/blob/8ace9af200dda345be3feab2b63a0a3277246243/integrations/formspree/package.json#L11)).

taranvohra: I think it's probably because we don't mark `devDeps` as **external** during build process 🤔 

https://github.com/GitbookIO/integrations/blob/8ace9af200dda345be3feab2b63a0a3277246243/packages/cli/src/build.ts#L33

taranvohra (approved):

PR #201, opened 2023-08-02
use a JSON encoded state for OAuth

taranvohra:
Now we can also store access tokens in space-installation config if the request was initiated via a space-installation public url.


dstotijn (approved):
LGTM! ✨

PR #206, opened 2023-08-05
Add terraform integration

dstotijn:


integrations/terraform/src/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+import { TerraformInstallationConfiguration, TerraformRuntimeContext } from './configuration';
+import { generateJwtSecret, parseJwtSecret } from './token';
+
+const tokenRegExp = /^Bearer\s+(\S+)$/i;
+
+const tokenText = createComponent<{}, {}, void, TerraformRuntimeContext>({
+    componentId: 'jwt',
+    async render(element, context) {
+        element.setCache({ maxAge: 30 });
```
dstotijn, line 13: Note to self: Remove before merging.

integrations/terraform/gitbook-manifest.yaml
```diff
@@ -0,0 +1,27 @@
+name: terraform
+title: Terraform
+organization: gitbook
+visibility: private
+description: An integration to support syncing Terraform resources with GitBook.
+script: src/index.tsx
+scopes:
```
dstotijn, line 7: These are just placeholder scopes for now; we'll need to review and add support for any missing ones.

integrations/terraform/src/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+const handleFetchEvent: FetchEventCallback<TerraformRuntimeContext> = async (request, context) => {
+    const { api } = context;
+    const authToken = (request.headers.get('Authorization') || '').match(tokenRegExp)?.[1];
+    if (!authToken) {
+        return new Response('Unauthorized', { status: 401 });
+    }
+
+    // First, parse the installation ID from the JWT *without* verifiying the
+    // signature, which we need to do so we can fetch the sign secret from the
+    // installation configuration.
```
dstotijn, line 46: The benefit of this approach is that the Terraform provider can just use a fixed URL to obtain a token; e.g. no need to separately configure an installation ID.

PR #224, opened 2023-08-29
Pass waitUntil to Handler

scazan:
- pass FetchEvent down to enable waitUntil
- bump version

### DIFF
--- a/packages/runtime/src/context.ts
+++ b/packages/runtime/src/context.ts
@@ -23,6 +23,11 @@ export interface RuntimeContext<Environment extends RuntimeEnvironment = Integra
      */
     api: GitBookAPI;
 
+    /**
+     * Wait for any pending promises to complete before finishing the execution.
+     * Use this function for any side-effects in your handlers that are not awaited, but must finish before
+     * the execution context is closed.
+     */
     waitUntil: FetchEvent['waitUntil'];
 }
 
--- a/packages/runtime/src/context.ts
+++ b/packages/runtime/src/context.ts
@@ -22,6 +22,8 @@ export interface RuntimeContext<Environment extends RuntimeEnvironment = Integra
      * Authenticated client to the GitBook API.
      */
     api: GitBookAPI;
+
+    event: FetchEvent;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


packages/runtime/src/integrations.ts
```diff
@@ -61,7 +61,8 @@ export function createIntegration<Context extends RuntimeContext = RuntimeContex
             const event = JSON.parse(formData.get('event') as string) as Event;
             const fetchBody = formData.get('fetch-body');
             const context = createContext(
-                JSON.parse(formData.get('environment') as string) as IntegrationEnvironment
+                JSON.parse(formData.get('environment') as string) as IntegrationEnvironment,
+                ev
```
emmerich, line 65: Don't put it on the context, prefer passing it directly to the handler on line 78:

```
await definition.fetch(request, ev, context)
```

This keeps a cleaner scope between a context (potentially long-lived across multiple requests) and a request/FetchEvent (different per request)

emmerich (changes requested):

packages/runtime/src/integrations.ts
scazan: shouldn't this be  `await definition.fetch(request, context, ev)` to prevent breaking changes to current integrations?

emmerich: aha I've read the Slack thread and I'm more up-to-date now.

In that case I'd recommend as you've done using the context but just passing the `waitUntil` function, so calling code could do: `context.waitUntil(() => {})`

emmerich: Main reason is we want to avoid something doing `context.ev.request.url` or something

scazan: ok, updated!

packages/runtime/src/context.ts
```diff
@@ -22,6 +22,8 @@ export interface RuntimeContext<Environment extends RuntimeEnvironment = Integra
      * Authenticated client to the GitBook API.
      */
     api: GitBookAPI;
+
+    waitUntil: FetchEvent['waitUntil'];
```
taranvohra, line 26: I'm not sure if overall this is a good structure on how we should be exposing parts of the event's request 🤔

emmerich: Why not?

emmerich, line 26:
```suggestion
    /**
     * Wait for any pending promises to complete before finishing the execution.
     * Use this function for any side-effects in your handlers that are not awaited, but must finish before
     * the execution context is closed.
     */
    waitUntil: FetchEvent['waitUntil'];
```

As this is a public API we should document it

emmerich (approved):

PR #228, opened 2023-09-01
Switch command handler to routing patter to allow for url verification

scazan:
This changes the events handler over to the routing pattern used in other places. It allows us to handle the url_verify challenges from slack when you switch the events endpoint (which we need to do when we point it at different environments such as staging).

### DIFF
--- a/integrations/slack/src/handlers/lens.ts
+++ b/integrations/slack/src/handlers/lens.ts
@@ -1,6 +1,6 @@
 import { Logger } from '@gitbook/runtime';
 
-import { queryLens } from '../actions/queryLens'; // eslint-disable-line
+import { queryLens } from '../actions/queryLens'; // eslint-disable-line import/no-internal-modules
 import type { SlashEvent } from '../commands';
 import { SlackRuntimeContext } from '../configuration';
 import { stripBotName } from '../utils';
--- a/integrations/slack/src/handlers/lens.ts
+++ b/integrations/slack/src/handlers/lens.ts
@@ -1,9 +1,9 @@
 import { Logger } from '@gitbook/runtime';
 
-import { queryLens } from '../actions/queryLens'; // eslint-disable-line import/no-internal-modules
-import type { SlashEvent } from '../commands';
+import { queryLens } from '../actions';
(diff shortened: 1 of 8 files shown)


integrations/slack/src/handlers/lens.ts
```diff
@@ -1,8 +1,9 @@
 import { Logger } from '@gitbook/runtime';
 
-import { queryLens } from '../actions/queryLens';
+import { queryLens } from '../actions/queryLens'; // eslint-disable-line import/no-internal-modules
```
valentin0h, line 3: To get rid of the disable line and in case there's a `actions/index.ts`.

import { queryLens } from '../actions';

scazan: ah great. I was wondering why we had this.

scazan: ah right, that causes a conflict between `actions/` and `actions.ts` which is why it reaches here. will try to resolve that

integrations/slack/src/router.ts
```diff
@@ -68,12 +68,23 @@ export const handleFetchEvent: FetchEventCallback = async (request, context) =>
     );
 
     // event triggers, e.g app_mention
-    router.post('/events', verifySlackRequest, acknowledgeSlackRequest);
+    router.post(
+        '/events',
+        verifySlackRequest,
+        createSlackEventsHandler(
```
valentin0h, line 74: It's slightly confusing that we handle both `/events` and `/events_task`. Maybe for another PR, but what do you think of something like this

```
 router.post(
        '/events',
        verifySlackRequest,
        acknowledgeSlackRequest // which can handle url verification
)
        
```

```
 router.post(
        '/events_task',
        verifySlackRequest,
        createSlackEventsHandler(...) // concerned with handling logic rather than ack
```

valentin0h (approved):

integrations/slack/src/router.ts
scazan: hmm... I had the the opposite thought. thought it was confusing to be handling routing in two places in two different ways. I think doing that would hide the url_verification rather than follow the same pattern.

PR #235, opened 2023-09-05
Allow Slash Events in DMs

scazan:
This switches most replies to use the response_url as it is more consistent across different messaging contexts. This enables the ability to use the slack commands from within a DM for instance which was previously not possible.
It does introduce some complexity as we don't always want to use response_url in every context and that is commented for now so we can get it pushed out.

- ensure that we are sharing without it being ephemeral
- add comment for odd boolean

### DIFF
--- a/integrations/slack/src/handlers/actions.ts
+++ b/integrations/slack/src/handlers/actions.ts
@@ -15,12 +15,16 @@ export function createSlackActionsHandler(
     return async (request, context) => {
         const actionPayload = await parseActionPayload(request);
 
-        const { actions, container, channel, team, user } = actionPayload;
+        const { actions, container, channel, channel_name, team, user, response_url } =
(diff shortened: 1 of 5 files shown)


integrations/slack/src/handlers/actions.ts
```diff
@@ -35,6 +42,8 @@ export function createSlackActionsHandler(
                     // pass user if exists
                     ...(user.id ? { userId: user.id } : {}),
 
+                    // if we have a response_url, we can reply safely using that
+                    responseUrl: dontUseResponseUrl ? null : response_url,
```
valentin0h, line 46:
```suggestion
                    responseUrl: dontUseResponseUrl ? undefined : response_url,
```

valentin0h (approved):

PR #248, opened 2023-10-03
Add Captures category to Slack integration

clairechabas:
This is for the integrations modal.

Contributes to CUR-988


taranvohra (approved):
Looks good 👍

Issue #240, opened 2023-09-09
Mermaid click feature support


jeffabailey:
Using [the Mermaid click feature](https://mermaid.js.org/syntax/flowchart.html#interaction) with the _blank option within the Mermaid integration results in undesirable behavior.

See the following video to see steps, and this page to try it out.

**Page:** https://jeffs-organization-3.gitbook.io/managing-inner-source-projects/~/changes/NdT1gsJCKqwbVRhqQRZF/use_gqm

**Video:** https://github.com/GitbookIO/integrations/assets/776901/2d1d6d4e-4116-4384-8b11-3e08031cd875

rrrutledge:
The Mermaid graph renders in a small iframe, and click events meant to take you to some full-page view stay rendered within that iframe.  We'd rather have the click targets take you to a full-page view of the target.

addisonschultz:
Can you try and add `_top` to the link you're trying to open?

jeffabailey:
Hi Addison, thanks for the response.

Yes, I tried using _top along _blank and _parent as described here:

https://mermaid.js.org/syntax/flowchart.html#interaction

Here's an updated version of the page demonstrating the issue more directly.

https://jeffs-organization-3.gitbook.io/managing-inner-source-projects/introduction-1/goals/use_gqm

Underlying mermaid code.

```
flowchart LR
    blank-->top
    top-->parent 
    
    click blank "https://www.github.com" _blank
    click top "https://www.github.com" _top
    click parent "https://www.github.com" _parent
```

PR #252, opened 2023-10-03
Mermaid - Add securityLevel loose to open links in a new tab

rodrcastro:
New links can be opened in a new tab based on the [Mermaid documentation](https://mermaid.js.org/syntax/flowchart.html#interaction). This requires the security level to be set to loose.

This PR adds the security level loose to allow for links to be opened in a new tab.

### DIFF
--- a/integrations/mermaid/src/index.tsx
+++ b/integrations/mermaid/src/index.tsx
@@ -70,7 +70,7 @@ export default createIntegration({
                 <body>
                     <script type="module">
                         import mermaid from 'https://cdn.jsdelivr.net/npm/mermaid@10/dist/mermaid.esm.min.mjs';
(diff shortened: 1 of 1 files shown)


taranvohra:
Hm, it's a bit risky because `securityLevel: 'loose'` can allow JS callbacks to be executed which makes it prone to XSS attacks. I am not really sure if we should offer it as a default behaviour. @SamyPesse any thoughts?

addisonschultz:
PR related to issue logged: https://github.com/GitbookIO/integrations/issues/240

addisonschultz:
@rodrcastro checking back on this - is this something that users are still asking about? I'm cleaning up some PRs in this repo

rodrcastro:
@addisonschultz Nope, they didn't come back for this.

PR #272, opened 2023-10-09
Slack: Ignore bot messages in thread

valentin0h:


integrations/slack/src/actions/saveThread.ts
```diff
@@ -154,16 +154,29 @@ async function createMessageThreadCapture(slackEvent, context: SlackRuntimeConte
 
     const capture = startCaptureRes.data;
 
-    const events = messages.map((message) => {
-        const { text, ts, thread_ts } = message;
+    const events = messages
+        .filter((message) => {
+            // ignore messages in thread from any bot (todo: potentially limit to gitbook only)
+            if (message.bot_id) {
```
scazan, line 160: excellent

integrations/slack/src/utils.ts
```diff
@@ -100,3 +102,20 @@ export function getActionNameAndType(actionId: string) {
...
+export function isSaveThreadMessage(message: string) {
+    const tokens = message.split(' ');
+
+    if (tokens.length > 3) {
+        return false;
+    }
+
+    const [first, second] = tokens;
+
+    // `save` or `save this` or `please save`
```
scazan, line 115: I like this "fuzziness" . for the future we could also just match on /.+save.+/ in the first few words. This doesn't really account for "Our API is now at v2 @GitBook save" which could also happen.

scazan (approved):

PR #291, opened 2023-11-10
[draft] Add tldraw integration

TodePond:
This PR adds a simple integration for tldraw, ahead of our call next week.

There are more examples on [this page right here](https://tldraw.gitbook.io/), that we can also talk through.

---

The example in this PR is a custom block for when you paste a tldraw multiplayer project. It can also be accessed by typing `/tldraw`. It embeds the multiplayer project in a modal, similar to some of the other integrations. You can paste a public, read-only, or snapshot board. And you can try it [here](https://tldraw.gitbook.io/tldraw-integration/).

**Future work** can include thumbnail support, an embedded editor, and displaying just an exported image in 'view mode'. (I'm looking forward to speaking with you about this).

### DIFF
--- a/.changeset/silver-spies-change.md
+++ b/.changeset/silver-spies-change.md
@@ -1,0 +1,5 @@
+---
+'tldraw-gitbook-block': minor
+---
+
+Add tldraw integration


addisonschultz:
Hi @TodePond! 

Thanks for putting this together, it's looking great - and the code is working as expected as well. Just a few requests, and some small questions on the UX of the integration. 

The first set of small things - in order for us to verify and list this integration in our marketplace, we'll need some assets and more metadata (preview images, descriptions, summaries, external links) to make the listing a bit more friendly for anyone discovering the integration. We have more info on what you can/should add here! 

https://developer.gitbook.com/marketplace/submit-your-app-for-review#prepare-assets

The second thing is more of the UX of the integration. I see in the code you've prepared 2 different blocks (and saw that in your demo / demo file as well 😄). 

My question is what the default behavior should be - is the intended default to be the modal/click to open modal? 

I do think some of the other workflows/versions could also be useful - like the canvas version in edit mode - and the static version in the published version. 

That being said, it's totally up to you how you'd like to have this app work - but maybe it might be worth exploring adding a configuration option in the block itself to let the user choose what version/type they want?

You can add controls for a block directly on the element: https://developer.gitbook.com/integrations/contentkit/reference/reference#layout

We have an example of that in the GitHub Files integration as well: https://github.com/GitbookIO/integrations/blob/0634853a6779532099ad86c3da8381123967b09f/integrations/github-files/src/index.tsx#L79

Let me know what you think, and if you can re-publish with some assets, and we'll go from there!

TodePond:
Hey, revisiting this now! We suddenly became very busy (with AI demos) shortly after opening this PR.
Thank you for your feedback - that's very helpful. I hope to dive back in at some point soon, and published a finalised version.

PR #292, opened 2023-11-13
Don't allow external channels to use the GitBook bot

scazan:
- don't respond to requests in external channels
- don't allow external channels in app mentions or messages
- update comment


integrations/slack/src/handlers/handlers.ts
```diff
@@ -73,9 +74,10 @@ export async function messageEventHandler(eventPayload: any, context: SlackRunti
 export async function appMentionEventHandler(eventPayload: any, context: SlackRuntimeContext) {
     // pull out required params from the slashEvent for queryLens
     const { type, text, bot_id, thread_ts, channel, user, team } = eventPayload.event;
+    const isExternalChannel = eventPayload.is_ext_shared_channel;
 
     // check for bot_id so that the bot doesn't trigger itself
-    if (['message', 'app_mention'].includes(type) && !bot_id) {
+    if (['message', 'app_mention'].includes(type) && !bot_id && !isExternalChannel) {
```
jpreynat, line 80: Looks like `bot_id || isExternalChannel` (or the reverse `!bot_id && !isExternalChannel`) condition is used in multiple place to accept/reject an event.
Might be worth making this a function accepting an `eventPayload` and return whether it should be processed or not, so we can make sure that we don't forget a condition in any of the checks.

jpreynat (approved):

SamyPesse (approved):

PR #303, opened 2023-11-21
Slack notification, use semantic changes

sdirosa:
Uses the new space revision semantic changes API to build a detailed summary of what changed, when receiving a `space_content_updated` event callback, and use that for the Slack notification text.

### DIFF
--- a/integrations/slack/src/index.ts
+++ b/integrations/slack/src/index.ts
@@ -48,7 +48,7 @@ const handleSpaceContentUpdated: EventCallback<
      *
      *    Content of *Space* has been updated.
      *
-     *    Summary of Changes:
+     *    Summary of changes:
      *    • New pages: Page1, Page2
      *    • Modified pages: Page3
      *    • Deleted pages: Page4, Page5
@@ -109,7 +109,7 @@ const handleSpaceContentUpdated: EventCallback<
         editedFiles.length > 0 ||
         deletedFiles.length > 0
     ) {
-        notificationText += '\n\nSummary of Changes:';
+        notificationText += '\n\nSummary of changes:';
 
         if (createdPages.length > 0) {
             notificationText += `\n• New pages: ${createdPages.join(', ')}`;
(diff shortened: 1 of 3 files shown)


integrations/slack/src/index.ts
```diff
@@ -26,8 +27,115 @@ const handleSpaceContentUpdated: EventCallback<
...
+                break;
+            case RevisionSemanticChangeType.FileDeleted:
+                deletedFiles.push(change.file.name);
+                break;
+            default:
+                break;
+        }
+    });
+
+    let notificationText = `Content of *${space.title || 'Space'}* has been updated.`;
```
sdirosa, line 99: Samy suggested to remove links as we don't know whether the recipient of the notification has access to the space or not. Up to discussion. Easy to reintroduce if we prefer that.

integrations/slack/src/index.ts
```diff
@@ -26,8 +27,117 @@ const handleSpaceContentUpdated: EventCallback<
...
+                deletedFiles.push(change.file.name);
+                break;
+            default:
+                break;
+        }
+    });
+
+    let notificationText = `Content of *<${space.urls.app}|${
+        space.title || 'Space'
+    }>* has been updated.`;
```
sdirosa, line 101: A note here. Samy pointed out that linking to spaces may be a issue for some recipients as they may actually not have access to them. I left it as-is for now, but we could just print the space title here, without linking it.

scazan: I'm wondering if there is any downside to them not having access when they click through. Won't it indicate that they don't have access (prompting them to either ask and admin or informing them that they can't actually look at it)? Feels odd that there is no link there when I see it in action.

scazan: I'm gonna file a Linear issue for this topic. I don't think it should hold up pushing it out.

sdirosa: I think it's a hard one to solve. We can't determine here who is going to receive the notification. I think from a UX perspective what you suggest is what makes the most sense. Accepting that some users may receive the notification and get denied and have some nice UI for it.

PR #304, opened 2023-11-21
Fix bug where CLI fails when installing dependencies on Windows machines

addisonschultz:
This PR should fix the issue where initializing a new GitBook project fails due to the engine installing dependencies not installing them correctly on Windows machines.


addisonschultz:
* Not yet tested properly on Windows machine
* Not deployed, can go out with a future deploy

PR #335, opened 2024-01-04
Send event timestamp when triggering Git sync imports and exports

spastorelli:
Following https://github.com/GitbookIO/gitbook-x/pull/9118, updating the Git integration to send the event timestamp when triggering imports and exports via the API.

### DIFF
--- a/integrations/github/src/sync.ts
+++ b/integrations/github/src/sync.ts
@@ -43,10 +43,18 @@ export async function triggerImport(
 
         /** Whether the git info should be updated on the space */
         updateGitInfo?: boolean;
+
+        /**
+         * The date/time of the event that triggers the import.
+         *
+         * This is to help ensures that Git sync import and export operations are executed
+         * in the same order on GitBook and on the remote repository.
+         */
+        eventCreatedAt?: Date;
     } = {}
 ) {
     const { api } = context;
-    const { force = false, updateGitInfo = false, standalone } = options;
+    const { force = false, updateGitInfo = false, standalone, eventCreatedAt } = options;
 
     const config = getSpaceConfigOrThrow(spaceInstallation);
     assertIsDefined(config.branch, { label: 'config.branch' });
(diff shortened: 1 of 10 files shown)


integrations/github/src/sync.ts
```diff
@@ -43,10 +43,18 @@ export async function triggerImport(
...
         /** Whether the git info should be updated on the space */
         updateGitInfo?: boolean;
+
+        /**
+         * The date/time of the event that triggers the import.
+         *
+         * This is to help ensures that Git sync import and export operations are executed
+         * in the same order on GitBook and on the remote repository.
+         */
+        eventCreatedAt?: Date;
```
jpreynat, line 53: I think that `eventTimestamp` would be a slightly better name 🙃

integrations/github/src/webhooks.ts
```diff
@@ -147,6 +151,10 @@ export async function handlePullRequestEvents(
                         standalone: {
                             ref: headRef,
                         },
+                        eventCreatedAt:
```
jpreynat, line 154: I'm wondering if we should add it here.
If I'm not wrong, those are for PR previews, so they are standalone and don't need to be ordered.
CC @taranvohra Wdyt?

integrations/gitlab/src/webhooks.ts
```diff
@@ -171,6 +194,10 @@ export async function handleMergeRequestEvent(
                         standalone: {
                             ref: sourceRef,
                         },
+                        eventCreatedAt:
```
jpreynat, line 197: Same here, I don't think that it's necessary for merge requests events

jpreynat (approved):

integrations/github/src/sync.ts
spastorelli: Yep agree, renamed it, thanks.

integrations/github/src/webhooks.ts
taranvohra: Yeah let's avoid it for PR previews

integrations/gitlab/src/webhooks.ts
taranvohra: Yep!

integrations/github/src/webhooks.ts
spastorelli: Removed it for previews.

PR #368, opened 2024-01-21
Add VA Okta Integration

vibhanshub:


integrations/va-okta/src/index.tsx
```diff
@@ -0,0 +1,284 @@
...
+        const clientId = environment.spaceInstallation?.configuration.client_id;
+        const location = event.location ? event.location : '';
+
+        const url = new URL(`https://${oktaDomain}/oauth2/default/v1/authorize`);
+        url.searchParams.append('client_id', clientId);
+        url.searchParams.append('response_type', 'code');
+        url.searchParams.append('redirect_uri', `${installationURL}/visitor-auth/response`);
+        url.searchParams.append('response_mode', 'query');
+        url.searchParams.append('scope', 'openid');
+        url.searchParams.append('state', `state-${location}`);
```
taranvohra, line 274: Why does it have to be prefixed with `state-`

taranvohra (approved):

integrations/va-okta/src/index.tsx
vibhanshub: Woops, somehow missed this comment. It's because Okta does not allow empty string as `state`, whereas Auth0 and Azure allow empty string as `state`.

PR #404, opened 2024-02-14
Improve logging for VA integrations

vibhanshub:


integrations/va-auth0/src/index.tsx
```diff
@@ -225,9 +225,19 @@ const handleFetchEvent: FetchEventCallback<Auth0RuntimeContext> = async (request
...
-                        });
+                        logger.debug(
+                            `Did not receive access token. Error: ${(resp && resp.error) || ''} ${
+                                (resp && resp.error_description) || ''
+                            }`
+                        );
+                        return new Response(
+                            `Error: No Access Token found in response from Auth0 ${
+                                (resp && resp.error) || ''
+                            } ${(resp && resp.error_description) || ''}`,
```
scazan, line 236: Have you verified that error.description won't possibly leak any keys into the logs?

vibhanshub: These errors are returned by the auth providers, so they're unlikely to contain any keys. However, there is no way to confirm that. At least I can't think of a way to confirm that

taranvohra: I agree with @scazan. We should not surface errors directly as it could contain sensitive information.

taranvohra (dismissed):

integrations/va-auth0/src/index.tsx
vibhanshub: @taranvohra but is it fine to log them? And can we surface just response.error to the user (and not response.error_description)

taranvohra: Logs are fine because only the integration owner can see them.

vibhanshub: Fixed!

scazan (approved):

PR #445, opened 2024-04-10
Add AWS Cognito VA integration

vibhanshub:


integrations/cognito/gitbook-manifest.yaml
vibhanshub, line 1: previewImage will be added when Marketing provides me with one.

Visibility will be set to public and organization changed to GitBook when the PR is approved

integrations/cognito/gitbook-manifest.yaml
```diff
@@ -0,0 +1,31 @@
+name: cognito
+title: AWS Cognito
+icon: ./assets/icon.png
+previewImages:
+    - ./assets/preview.png
+description: Control who has access to your published documentation with AWS Cognito
+
+visibility: private
```
emmerich, line 8: Maybe publish as private until we have full marketing assets then public + verify it once it's ready.

integrations/cognito/gitbook-manifest.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+    This integration lets you control access to your published content as determined by AWS Cognito.
+
+    # Configure
+    Install this integration on a space and then populate the configuration screen with the details of your AWS Cognito application and AWS Cognito account.
+    You can then open the Share menu, publish the space with Visitor Authentication, choose this integration as the authentication backend, and hit Save.
+
+    Your space is now published with Visitor Authentication using AWS Cognito.
+categories:
+    - other
```
emmerich, line 28: Side thought: it would be good to introduce a visitor auth category

integrations/cognito/src/index.tsx
```diff
@@ -0,0 +1,357 @@
...
+    };
+    siteInstallation: {
+        configuration?: CognitoSiteOrSpaceInstallationConfiguration;
+    };
+};
+
+export type CognitoAction = { action: 'save.config' };
+
+const getDomainWithHttps = (url: string): string => {
+    if (url.startsWith('https://') || url.startsWith('http://')) {
```
emmerich, line 46: Should you convert all traffic to `https://` ?

emmerich (approved):

integrations/cognito/gitbook-manifest.yaml
vibhanshub: We have all the marketing assets. But I agree with keeping it private until we've tested it and it looks good!

integrations/cognito/src/index.tsx
vibhanshub: Yes, there is no reason that the AWS Cognito user domain has the `http://` protocol. So we convert all domains without a protocol to `https://`.

but if the user, for some reason, specifies the `http://` protocol, we keep it. But it's unlikely to work. I am only keeping this here because of some edge case for the need of an `http://` protocol that I am unaware of (could not find anything in the Cognito docs)

taranvohra: I don't think we should keep it. We should force all communication to be https

taranvohra (approved):

PR #453, opened 2024-04-23
Add site configuration to manifest of integration that can be installed on a site

spastorelli:


taranvohra (approved):

addisonschultz:
Is this a requirement now?

spastorelli:
> Is this a requirement now?

Yes for the time being for integrations that are running on published content (e.g inject scripts/va etc...) during the transition until we migrate fully to sites. After that integrations that can be installed on sites will just need one config (i.e site configuration)

addisonschultz:
@spastorelli  Okay - I need to communicate this with a few teams I've allowed to use this scope for analytics solutions, for custom solutions for their teams etc. 

So they need to add this to the configuration, and the site ID is the space id?

```
    site:
        properties:
            site_id:
                type: string
                title: Site ID
                description: Look for this in your Fathom analytics account
        required:
            - site_id
```

PR #467, opened 2024-05-02
Shorten description for Slack integration

addisonschultz:
Small description change


.changeset/old-rabbits-impress.md
taranvohra, line 1: We need only 1 changeset file. Can probably delete this one

taranvohra (approved):

PR #547, opened 2024-09-17
Normalize GitHub config project directory on save to make it relative

jpreynat:
Because using an absolute path value (ie starting by `/` like `/path/to/project`) doesn't work, we make sure that paths are always relative.


integrations/github/src/installation.ts
taranvohra, line 1: Can you add the same for `gitlab`?

taranvohra (approved):
Nice! can you also add changesets (for github & gitlab) so when we merge we can have a release PR to trigger a new deployment for these integrations

integrations/github/src/installation.ts
jpreynat: Thanks, I forgot it indeed. And done 👍

PR #561, opened 2024-10-08
Google Analytics updated configuration

valentin0h:


integrations/googleanalytics/gitbook-manifest.yaml
```diff
@@ -53,10 +51,4 @@ categories:
...
 configurations:
     site:
-        properties:
-            tracking_id:
-                type: string
-                title: Tracking ID
-                description: Look for this in your Google Analytics account.
-        required:
-            - tracking_id
+        componentId: config
```
clairechabas, line 54: Just curious: why are we replacing the tracking id with a component id? 

Could we add a title and description to it to clarify its role and use?

clairechabas (commented):
Just took a look and asked a question but can't approve since I'm not familiar with integrations configuration in this repo.

integrations/googleanalytics/gitbook-manifest.yaml
valentin0h: Have a look [here](https://github.com/GitbookIO/gitbook-x/pull/14364) for more context. 

We're replacing the way we render the configuration. From controlled by declarative + controlled by gitbook-x to controlled by the integration completely.

valentin0h: [This](https://gitbook.slack.com/archives/C07HKUX4JPN/p1727776084802119) as well.

PR #568, opened 2024-10-10
Remove space code from VA integrations

vibhanshub:


integrations/oidc/src/index.tsx
```diff
@@ -243,38 +229,41 @@ const configBlock = createComponent<OIDCProps, OIDCState, OIDCAction, OIDCRuntim
...
  */
 async function getPublishedContentUrls(context: OIDCRuntimeContext) {
-    const organizationId = context.environment.installation?.target?.organization;
-    const siteOrSpaceInstallation =
-        context.environment.siteInstallation ?? context.environment.spaceInstallation;
-    const publishedContentData =
-        'site' in siteOrSpaceInstallation
-            ? await context.api.orgs.getSiteById(organizationId, siteOrSpaceInstallation.site)
-            : await context.api.spaces.getSpaceById(siteOrSpaceInstallation.space);
+    const organizationId = context.environment.installation?.target?.organization!;
```
valentin0h, line 235: Can we assert this as well?

integrations/oidc/src/index.tsx
```diff
@@ -351,9 +339,12 @@ const handleFetchEvent: FetchEventCallback<OIDCRuntimeContext> = async (request,
                         );
                     }
                 } else {
-                    return new Response('Error: Either ClientId or Client Secret is missing', {
-                        status: 400,
-                    });
+                    return new Response(
+                        'Error: Either ClientId or Client Secret or Access Token Endpoint is missing',
```
valentin0h, line 343: We can easily make the message more specific with a conditional. I think it's worth it, wdyt?

integrations/va-auth0/src/index.tsx
```diff
@@ -52,58 +50,44 @@ const getDomainWithHttps = (url: string): string => {
...
     componentId: 'config',
     initialState: (props) => {
-        const siteOrSpaceInstallation = props.siteInstallation ?? props.spaceInstallation;
+        const siteInstallation = props.siteInstallation;
         return {
-            client_id: siteOrSpaceInstallation?.configuration?.client_id?.toString() || '',
-            issuer_base_url:
-                siteOrSpaceInstallation?.configuration?.issuer_base_url?.toString() || '',
-            client_secret: siteOrSpaceInstallation?.configuration?.client_secret?.toString() || '',
+            client_id: siteInstallation?.configuration?.client_id?.toString() || '',
```
valentin0h, line 55: Do we need the `.toString()`?

valentin0h (commented):
Looks good but I'm wondering if we need to remove more stuff in the gitbook manifest, like the `space.visitor:auth` scope. Or is that still needed?

integrations/oidc/src/index.tsx
vibhanshub: All these things are configured from the same screen, so even if make the message more specific, the user will see the other items as well (missing or not).

I don't think it's worth it tbh since when any of these is missing, the user will see all the other items as well

valentin0h (approved):

PR #572, opened 2024-10-11
Migrate Segment integration to react to site_view events

spastorelli:
This PR aims to migrate the Segment integration to react to `site_view` integration events.

### DIFF
--- a/integrations/segment/gitbook-manifest.yaml
+++ b/integrations/segment/gitbook-manifest.yaml
@@ -11,7 +11,6 @@ visibility: public
 organization: gitbook
 script: ./src/index.ts
 scopes:
-    - space:views:read
     - site:views:read
 summary: |
     # Overview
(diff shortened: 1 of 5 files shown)


integrations/segment/gitbook-manifest.yaml
```diff
@@ -12,6 +12,7 @@ organization: gitbook
 script: ./src/index.ts
 scopes:
     - space:views:read
```
SamyPesse, line 14: Can we remove this one?

integrations/segment/gitbook-manifest.yaml
```diff
@@ -33,6 +34,14 @@ summary: |
...
     - analytics
 configurations:
+    site:
+        properties:
+            write_key:
+                type: string
+                title: Write Key
+                description: Key provided by Segment to identify the data source.
+        required:
+            - write_key
```
SamyPesse, line 44: Does it mean that users will have to reconfigure or can we run a migration?

SamyPesse (approved):
Looks good, but you need to run `bun run changeset` for this package

integrations/segment/gitbook-manifest.yaml
spastorelli: yeah we will have to run a migration to migrate space level installation to their sites

PR #629, opened 2024-11-21
Add support to VA Auth0 integration to include claims from backend in VA token

spastorelli:


integrations/va-auth0/src/index.tsx
```diff
@@ -22,6 +22,7 @@ type Auth0SiteInstallationConfiguration = {
     client_id?: string;
     issuer_base_url?: string;
     client_secret?: string;
+    include_claims_in_va_token?: boolean;
```
SamyPesse, line 25: Should we name this "enrich_session" or something simpler? the current name is very low level

integrations/va-auth0/src/index.tsx
```diff
@@ -89,83 +107,104 @@ const configBlock = createComponent<Auth0Props, Auth0State, Auth0Action, Auth0Ru
...
+                            }
+                            element={<codeblock content={VACallbackURL} />}
+                        />
+                    </vstack>
+                    <divider size="medium" />
+
+                    <markdown content="### Visitor authentication settings" />
+                    <input
+                        label="Include claims in JWT token"
+                        hint="Add user & custom claims from Auth0 backend to the JWT token to enrich the user's experience while navigating the site"
```
SamyPesse, line 187: This description should be revisited to discuss the goal, not what it does technically

integrations/va-auth0/src/index.tsx
```diff
@@ -213,78 +252,116 @@ const handleFetchEvent: FetchEventCallback<Auth0RuntimeContext> = async (request
...
+                    // Auth0 API application during the authorization request.
+                    // In this case, since we only request a token to verify authentication and generate the VA,
+                    // there is no valid API target to specify as audience.
+                    // As a result, we must call the /userinfo endpoint (an OIDC-compliant endpoint) to retrieve user claims.
+                    //
+                    // An alternative approach would be to request an ID Token using the Implicit Grant flow (with form post).
+                    // However, many authentication providers discourage this method in favor of the Authorization Code flow.
+                    // Additionally, some customers may disable the Implicit Grant flow in their Auth0 application.
+                    // Therefore, retrieving user data via the /userinfo endpoint is a more robust solution.
+                    const auth0TokenData = await auth0TokenResp.json<Auth0TokenResponseData>();
```
SamyPesse, line 311: Curious why we don't use a auth0 library or something

integrations/va-auth0/src/index.tsx
```diff
@@ -330,9 +407,22 @@ export default createIntegration({
...
         url.searchParams.append('response_type', 'code');
+        url.searchParams.append('scope', 'openid');
         url.searchParams.append('redirect_uri', `${installationURL}/visitor-auth/response`);
         url.searchParams.append('state', location);
 
         return Response.redirect(url.toString());
     },
 });
+
+function sanitizeJWTTokenClaims(claims: jwt.JwtPayload) {
```
SamyPesse, line 418: Why do we need to sanitize if it's done on the GitBook front as well?

integrations/va-auth0/src/index.tsx
```diff
@@ -330,9 +407,22 @@ export default createIntegration({
         const url = new URL(`${issuerBaseUrl}/authorize`);
         url.searchParams.append('client_id', clientId);
         url.searchParams.append('response_type', 'code');
+        url.searchParams.append('scope', 'openid');
```
SamyPesse, line 410: Is it only when we want to include the claims?

Is it going to impact existing users?

integrations/va-auth0/src/index.tsx
```diff
@@ -213,78 +252,116 @@ const handleFetchEvent: FetchEventCallback<Auth0RuntimeContext> = async (request
...
+                });
+                const accessTokenURL = `${issuerBaseUrl}/oauth/token/`;
+                const auth0TokenResp = await fetch(accessTokenURL, {
+                    method: 'POST',
+                    headers: { 'content-type': 'application/x-www-form-urlencoded' },
+                    body: searchParams,
+                });
+
+                if (!auth0TokenResp.ok) {
+                    const errorResponse = await auth0TokenResp.json<Auth0TokenResponseError>();
```
gregberge, line 286: This line could crash. Should be optional, you should do another try / catch around that.

integrations/va-auth0/src/index.tsx
```diff
@@ -213,78 +252,116 @@ const handleFetchEvent: FetchEventCallback<Auth0RuntimeContext> = async (request
...
+                    userInfo = await userInfoResp.json<Record<string, any>>();
+                }
+
+                try {
+                    const privateKey = context.environment.signingSecrets.siteInstallation;
+                    if (!privateKey) {
+                        return new Response('Error: Missing private key from site installation', {
+                            status: 400,
+                        });
+                    }
```
gregberge, line 339: Why is it in the try/catch?

integrations/va-auth0/src/index.tsx
```diff
@@ -213,78 +252,116 @@ const handleFetchEvent: FetchEventCallback<Auth0RuntimeContext> = async (request
...
+                    const jwtToken = await jwt.sign(
+                        {
+                            ...sanitizeJWTTokenClaims(userInfo || {}),
```
gregberge, line 342:
```suggestion
                            ...(userInfo ? sanitizeJWTTokenClaims(userInfo) : {}),
```

integrations/va-auth0/src/index.tsx
```diff
@@ -213,78 +252,116 @@ const handleFetchEvent: FetchEventCallback<Auth0RuntimeContext> = async (request
...
+                            exp: Math.floor(Date.now() / 1000) + 1 * (60 * 60),
```
gregberge, line 343: A comment would be nice

gregberge (approved):
I think the code could be splitted in several functions to make it clearer but good 👍

integrations/va-auth0/src/index.tsx
spastorelli: The previous code didn't use a library hence why I didn't look into using one. I haven't honestly looked into why we didn't use one before but it could have been aimed at reducing the integration bundle size or just that we didn't see the need to since we are only relying on a couple of endpoints, or maybe that none were found to be compatible with the Cloudflare worker runtime. Also I had very quick look at the Auth0 libraries and they seem to be focus on providing a middleware only but didn't look in details.

I'll look at refactoring the integration code to use a library in another PR.

integrations/va-auth0/src/index.tsx
spastorelli: Yes forgot to add the conditional inclusion sorry. Adding it. For Auth0 by default there is no request for consent by the user at login for only the `openid` scope but it is best to have this conditionally request that scope in case some customer setup have this require consent enabled.

integrations/va-auth0/src/index.tsx
spastorelli: You mean just the lines 334-339?  No particular reason. Moving it out.
For the rest (e.g jwt.sign) it can throw and best to catch and return a simpler error.

integrations/va-auth0/src/index.tsx
spastorelli: Extending the condition to be more defensive.

integrations/va-auth0/src/index.tsx
vibhanshub: @SamyPesse @spastorelli We don't use an auth0 library because Cloudflare works runtime only supports a few libraries, and Auth0 isn't one of them. It is also why for the JWT library we're using a lib that's built specifically for CF

PR #631, opened 2024-11-21
Add support to VA Azure integration to include claims from backend in VA token

spastorelli:


integrations/va-azure/src/index.tsx
```diff
@@ -204,79 +221,92 @@ const handleFetchEvent: FetchEventCallback<AzureRuntimeContext> = async (request
...
+                }
+
+                const azureTokenData = await azureTokenResp.json<AzureTokenResponseData>();
+                if (!azureTokenData.id_token) {
+                    return new Response('Error: No ID Token found in response from Azure', {
+                        status: 401,
+                    });
+                }
+
+                // Azure already include user/custom claims in the ID token so we can just decode it
```
emmerich, line 275: Okay, so this is why there's no configuration option for claims?

emmerich (approved):

integrations/va-azure/src/index.tsx
spastorelli: Indeed afaik Azure includes standard and custom claims in the ID token as long as the admin of the tenant decides to include/add them (by setting this up in their azure dash). So we just decode the claims directly from the token instead.

We should technically validate the tokens (like with the other providers too) against the audience (i.e `aud`) included in the claims hence why I added a TODO below. But this requires a bit more work (planning on tracking it) as you need to leverage JWKS since we don't have access directly to the signing key.

I think for now it's okay like this but need to be done at some point.

PR #657, opened 2024-12-12
Initialize Pendo Script

vibhanshub:
`pendo.initialize({})` needs to be called in the loaded script for pendo to work as expected. We are passing dummy data as we have no knowledge of the details expected of visitors.

This will initialize Pendo and make it work like Heap

### DIFF
--- a/packages/api/CHANGELOG.md
+++ b/packages/api/CHANGELOG.md
@@ -1,520 +1,526 @@
 # @gitbook/api
 
+## 0.82.0
+
+### Minor Changes
+
+- d7583bd: Bump API version to include latest sidebar customization settings
+
 ## 0.81.0
 
 ### Minor Changes
 
--   5c14848: Bump GitBook API.
+- 5c14848: Bump GitBook API.
 
 ## 0.80.1
 
 ### Patch Changes
 
--   879203a: Fix an issue where body data was not passed when using a streamed endpoint. It allows body data to be passed to POST endpoints that stream their response.
+- 879203a: Fix an issue where body data was not passed when using a streamed endpoint. It allows body data to be passed to POST endpoints that stream their response.
 
 ## 0.80.0
 
 ### Minor Changes
 
--   610c240: Bump @gitbook/api version for customization settings (tint)
+- 610c240: Bump @gitbook/api version for customization settings (tint)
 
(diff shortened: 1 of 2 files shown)


integrations/pendo/src/script.raw.js
```diff
@@ -21,5 +21,20 @@
         y.src = `${cdn}/agent/static/` + apiKey + '/pendo.js';
         z = e.getElementsByTagName(n)[0];
         z.parentNode.insertBefore(y, z);
+        y.onload = function () {
+            pendo.initialize({
```
spastorelli, line 25: Based on their documentation it looks like you may be able to call initialize with a special id to track visitor as anonymous:

> Call pendo.initialize() when your application first loads and then pass a special string ('VISITOR-UNIQUE-ID') to the Visitor ID. This generate a [Pendo anonymous Visitor ID](https://support.pendo.io/hc/en-us/articles/360032202751).
https://support.pendo.io/hc/en-us/articles/360031862272-Install-Pendo-on-a-single-page-web-application

Something like this
```suggestion
            pendo.initialize({
                visitor: {
                    id: 'VISITOR-UNIQUE-ID',
                 }
            });
```

have you tried this?

There also could be some side effect to tracking visitors as anonymous and never identifying them afterwards but not sure.

vibhanshub: @spastorelli I heard back from their support. This is what they said:

"I am not mega knowledgeable but the part you have right now doesn’t do anything, it’s just the declaration of what the Pendo snippet should do/behave etc. “pendo.initialize” is what actually runs a function to load the Pendo snippet on the page/app/etc so you need to have it in there. It doesn’t matter if you don’t have a login, this function is what loads Pendo in the thing, you can skip the visitor/account data, what will happen is that you will only get anonymous visitors/usage. Let me know if this helps!"

So I am thinking we can get away with `pendo.initialize({})`

vibhanshub: Wdyt?

spastorelli: Yep sounds good. I was only suggesting as this is something they mention in their docs. It's not very clear though but yeah if it works with empty object that's good.

spastorelli (approved):

PR #661, opened 2024-12-17
Increase the min length of state param to 10 in OIDC integration

vibhanshub:
Some IdPs enforce a min 8 character length for the state parameter. This changes the state param's min length from 6 to 10

### DIFF
--- a/.changeset/silver-clouds-bow.md
+++ b/.changeset/silver-clouds-bow.md
@@ -1,0 +1,5 @@
+---
+'@gitbook/integration-oidc': patch
+---
+
+Increase the min length of state to 10 chars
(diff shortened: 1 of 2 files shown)


gregberge (commented):
This change looks weird. You are talking about a char limit and changing a key.

vibhanshub:
@gregberge Yes, the `state` query param needs to be at least 8 chars long for some IdPs. Previously the state parameter being passed was just `state-`, which is 6 chars long when the `location` in VA is empty. Now, when the location is empty, we will pass `oidcstate-` as the value for the `state` query param

gregberge (approved):

PR #697, opened 2025-02-02
Update Slack with latest product changes

SamyPesse:
- [x] Fix integration using deprecated answer data
- [x] Remove capture


integrations/slack/src/handlers/handlers.ts
```diff
@@ -68,10 +68,10 @@ export async function messageEventHandler(eventPayload: any, context: SlackRunti
 }
 
 /**
- * Handle an Event request and route it to either GitBook Lens' query function or saveThread function.
+ * Handle an Event request and route it to either GitBook AskAI' query function or saveThread function.
```
valentin0h, line 71:
```suggestion
 * Handle an Event request and route it to either AskAI' query function.
```

integrations/slack/src/links.ts
```diff
@@ -43,7 +43,7 @@ export async function unfurlLink(event: LinkSharedSlackEvent, context: SlackRunt
     const installationApiClient = await api.createInstallationClient('slack', installation.id);
 
     // Resolve links to their content
-    const unfurls = {};
+    const unfurls: Record<string, any> = {};
```
valentin0h, line 46: Maybe this would be better?
```suggestion
    const unfurls: Record<string, object> = {};
```

valentin0h (approved):

integrations/slack/src/index.ts
```diff
@@ -165,58 +162,10 @@ const handleSpaceContentUpdated: EventCallback<
     });
 };
 
-/*
- * Handle content visibility being updated: send a notification on Slack.
- */
-const handleSpaceVisibilityUpdated: EventCallback<
```
addisonschultz, line 171: @SamyPesse Are we removing notifications for visibility updated? I guess we’re still keeping the content updated event though?

valentin0h: Space visibility isn't relevant anymore since the migration to sites.

PR #732, opened 2025-02-28
Use OpenAPI Operation block

gregberge:


SamyPesse (approved):
Approving but you will need to publish the API spec

PR #741, opened 2025-03-10
Support tag hierachy

gregberge:


integrations/openapi/src/parser/spec.ts
```diff
@@ -136,23 +138,83 @@ export function divideOpenAPISpecSchema(schema: OpenAPIV3.Document): OpenAPIGrou
...
+ * Get the OpenAPI tree from the schema.
+ * This will organize the pages by their tags and parent tags.
+ * It will also remove any pages that are marked as ignored in the spec.
+ * The root pages will be the ones that have no parent tag.
+ * The pages will be organized in a tree structure.
+ */
+export function getOpenAPITree(schema: OpenAPIV3.Document): OpenAPIPage[] {
+    const pages = getAllOpenAPIPages(schema);
+
+    if (!schema.tags) {
```
nolannbiron, line 154: Why not check `!schema.tags || !schema.tags.length` to avoid going further if there are no tags ?

gregberge: Because empty tags means you don't want to display any page, it's not the same meaning.

PR #755, opened 2025-03-17
Add command to CLI to publish OpenAPI API spec

gregberge:
- **Fix bun.lock**
- **Publish OpenAPI specification from CLI**

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -15,6 +15,7 @@ import { publishIntegration, unpublishIntegration } from './publish';
 import { authenticate, whoami } from './remote';
 import { tailLogs } from './tail';
 import { checkIntegrationBuild } from './check';
+import { publishOpenAPISpecification } from 'openapi/publish';
 
 program
     .name(Object.keys(packageJSON.bin)[0])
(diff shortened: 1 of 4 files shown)


packages/cli/src/cli.ts
```diff
@@ -115,6 +116,23 @@ program
         await checkIntegrationBuild();
     });
 
+const openAPIProgram = program.command('openapi').description('manage OpenAPI specifications');
```
taranvohra, line 119:
```suggestion
const openAPIProgram = program.command('openapi').description('Manage OpenAPI specifications');
```

packages/cli/src/cli.ts
```diff
@@ -115,6 +116,23 @@ program
...
+
+openAPIProgram
+    .command('publish')
+    .description('publish an OpenAPI specification')
```
taranvohra, line 123:
```suggestion
    .description('Publish an OpenAPI specification')
```

packages/cli/src/cli.ts
```diff
@@ -115,6 +116,23 @@ program
...
+    .argument('<file>', 'OpenAPI specification file path')
+    .requiredOption('-s, --spec <spec>', 'name of the OpenAPI specification')
```
taranvohra, line 125:
```suggestion
    .requiredOption('-s, --spec <spec>', 'Name of the OpenAPI specification')
```

packages/cli/src/cli.ts
```diff
@@ -115,6 +116,23 @@ program
...
+    .requiredOption('-o, --organization <organization>', 'organization to publish to')
```
taranvohra, line 126:
```suggestion
    .requiredOption('-o, --organization <organization>', 'Organization to publish to')
```

taranvohra (approved):

packages/cli/src/cli.ts
gregberge: There is no uppercase anywhere, that's why I kept the same style.

gregberge: If we do, we should do it for everything, else it's weird.

PR #763, opened 2025-03-18
CLI: Add string validation for integration names

talagluck:
Currently, it is possible to specify an integration name in the CLI package initialization flow that will show up as invalid when you try and publish the package. This PR adds a string validation step that uses the [same criteria as the API](https://github.com/GitbookIO/gitbook-x/blob/1d1355f3d5b7ddce6b64b279d1082f94c6f291a6/services/cloudrun/api/src/operations/integrations/publishIntegration.ts#L397) to make sure the string is valid.

I copied the validation functions directly since this repo is public, while the gitbook-x repo is private. I'm happy to update this if there's a better way to import that validation function directly here.


packages/cli/src/init.ts
```diff
@@ -11,6 +11,20 @@ import packageJSON from '../package.json';
 import { fileExists } from './files';
 import { DEFAULT_MANIFEST_FILE, writeIntegrationManifest } from './manifest';
 
+function validateURLSlug(slug: string): boolean {
+    return /^[a-z0-9]+(?:-[a-z0-9]+)*$/.test(slug);
+}
+
+function validateIntegrationName(name: string) {
+    const validName = /^[a-z0-9-]+$/g.test(name) as boolean;
```
SamyPesse, line 19: Why do we have 2 regex for the same thing?

packages/cli/src/init.ts
```diff
@@ -11,6 +11,20 @@ import packageJSON from '../package.json';
 import { fileExists } from './files';
 import { DEFAULT_MANIFEST_FILE, writeIntegrationManifest } from './manifest';
 
+function validateURLSlug(slug: string): boolean {
+    return /^[a-z0-9]+(?:-[a-z0-9]+)*$/.test(slug);
+}
+
+function validateIntegrationName(name: string) {
```
SamyPesse, line 18: Instead of creating a custom function, you should update `IntegrationManifestSchema` in `manifest.ts` to validate the `name` and reuse the same zod schema here

Something like:

```
const IntegrationNameSchema = z.string().regex(regex);
```

packages/cli/src/init.ts
```diff
@@ -22,6 +36,14 @@ export async function promptNewIntegration(dir?: string): Promise<void> {
             name: 'name',
             message: 'Name of the integration:',
             initial: path.basename(dir || process.cwd()),
+            validate: (value) => {
+                try {
+                    validateIntegrationName(value);
+                    return true;
+                } catch (error: unknown) {
+                    return (error as Error).message;
+                }
```
SamyPesse, line 45: Then this can become:

```
if (IntegrationNameSchema.safeParse(value)) {
  return true;
} else {
  return `Invalid integration name: ${value}, it must begin with an alphanumeric character and only contain alphanumeric characters and hyphens.`;
}
```

SamyPesse (approved):

PR #769, opened 2025-03-26
Fix CLI typescript errors

addisonschultz:
Updating the CLI generated template to fix CLI errors

### DIFF
--- a/packages/cli/src/init.ts
+++ b/packages/cli/src/init.ts
@@ -248,13 +248,13 @@ export function generateScript(project: { name: string }): string {
 
 export function generateTSConfig(): string {
     return detent(`
-            {
-                "extends": "@gitbook/tsconfig/integration.json",
-                "compilerOptions": {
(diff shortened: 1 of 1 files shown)


packages/cli/src/init.ts
```diff
@@ -246,7 +249,12 @@ export function generateScript(project: { name: string }): string {
 export function generateTSConfig(): string {
     return detent(`
         {
-            "extends": "@gitbook/tsconfig/integration.json"
+            "extends": "@gitbook/tsconfig/integration.json",
+            "compilerOptions": {
+                "lib": ["ES2015", "DOM"],
```
taranvohra, line 254: I think you can go higher `ESNext` or something. Let's also add `strict: true` as that has been the source of many unwanted errors in the past

taranvohra (approved):
You also need to add the changeset so that when this PR is merged you have a Release PR to release the new version of the CLI

PR #808, opened 2025-04-29
`@gitbook/adaptive` 

taranvohra:


packages/adaptive/src/client.ts
```diff
@@ -0,0 +1,53 @@
+import type { BucketClient } from '@bucketco/browser-sdk';
+import Cookies from 'js-cookie';
+
+const COOKIE_NAME = 'gitbook-visitor-public-bucket';
```
SamyPesse, line 4: I don't understand why the cookie name contains `bucket`. Is the idea to read ALL cookies that start with the prefix `gitbook-visitor-public-`?

packages/adaptive/src/client.ts
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Sets a client-side session cookie with the features enabled in the bucket client.
+ */
+export function withBucket(client: BucketClient): () => void {
```
SamyPesse, line 9: If this package has something as specific as a bucket integration, we should probably split into 2 packages:

- `@gitbook/adaptive` to write generic data, exports a function `writeGitBookCookie(prefix, data)`
- `@gitbook/adaptive-bucket` to export the `withBucket`

packages/adaptive/src/client.ts
taranvohra: Yes. GBO will read all cookies with that prefix.

packages/adaptive/src/client.ts
taranvohra: The packages are way too small to be split and I don't want to make that decision right now. What are your thoughts about 
 - `@gitbook/adaptive`: for generic data and functions
- `@gitbook/adaptive/bucket`: exports `withBucket`

SamyPesse: Sure, It can work, I'm not a big fan of this pattern usually, because it means a project will have to download the bucket client dependency, the launchdarkly one, etc, even if they don't use any (I know it'll not be bundled, but still).

> The packages are way too small and simple to be split and I don't want to make that decision right now

This is the whole point of NPM packages: they can be small

packages/adaptive/package.json
```diff
@@ -0,0 +1,35 @@
...
+        "@types/js-cookie": "^3.0.6",
+        "esbuild": "^0.25.3"
+    },
+    "scripts": {
+        "build": "bun esbuild.config.js && tsc --emitDeclarationOnly",
+        "typecheck": "tsc --noEmit"
+    },
+    "exports": {
+        ".": {
+            "import": "./dist/index.js",
```
gregberge, line 31: Why not .esm.js here?

packages/adaptive/package.json
```diff
@@ -0,0 +1,35 @@
+{
+    "name": "@gitbook/adaptive",
+    "description": "Utility to use adaptive content in GitBook",
+    "type": "module",
+    "version": "0.0.1",
+    "sideEffects": false,
+    "main": "dist/index.js",
+    "module": "dist/index.esm.js",
```
gregberge, line 8: It's deprecated, you should only use exports.

packages/adaptive/package.json
```diff
@@ -0,0 +1,35 @@
...
+    "unpkg": "dist/index.umd.js",
```
gregberge, line 9: We don't care.

packages/adaptive/package.json
```diff
@@ -0,0 +1,35 @@
...
+        "esbuild": "^0.25.3"
+    },
+    "scripts": {
+        "build": "bun esbuild.config.js && tsc --emitDeclarationOnly",
+        "typecheck": "tsc --noEmit"
+    },
+    "exports": {
+        ".": {
+            "import": "./dist/index.js",
+            "types": "./dist/index.d.ts"
```
gregberge, line 32: Add a default entry here.

packages/adaptive/package.json
```diff
@@ -0,0 +1,35 @@
...
+        "typescript": "^5.6.2",
+        "@bucketco/browser-sdk": "^3.1.2",
+        "@types/js-cookie": "^3.0.6",
+        "esbuild": "^0.25.3"
+    },
+    "scripts": {
+        "build": "bun esbuild.config.js && tsc --emitDeclarationOnly",
+        "typecheck": "tsc --noEmit"
+    },
+    "exports": {
```
gregberge, line 29: Also expose the `package.json` it's used by some package managers.

packages/adaptive/esbuild.config.js
gregberge, line 1: You should compile it with tsup, it's easier.

gregberge:
Also add `sideEffects: false` to the package.json.

packages/adaptive/esbuild.config.js
taranvohra: It's `esbuild` behind the scenes anyway 😄

Issue #822, opened 2025-05-19
Regex that captures URL in GitHub Files integration doesn't allow for '.' (dot) in repository name


boy-vd:
GitHub allows for the presence of dots in repository names, but the regex to capture and unfurl a URL to a file on GitHub doesn't account for this:

https://github.com/GitbookIO/integrations/blob/cb7ab97ddfd63c1ca10c654cc82f31e77c8e4a39/integrations/github-files/src/github.ts#L8-L12

This results in the GitHub Files integration not working for files in repositories like Next.js: https://github.com/vercel/next.js/ (instead of the contents of the file, the block in GitBook will display 'Not Found').

Solution is to change the regex for the group that captures the repository name slightly, from `([\w-]+)` to `([\w.-]+)`:
```
const splitGithubUrl = (url: string) => {
    const permalinkRegex =
        /^https?:\/\/github\.com\/([\w-]+)\/([\w.-]+)\/blob\/([a-f0-9]+)\/(.+?)#(.+)$/;
    const wholeFileRegex = /^https?:\/\/github\.com\/([\w-]+)\/([\w.-]+)\/blob\/([\w.-]+)\/(.+)$/;
    const multipleLineRegex = /^L\d+-L\d+$/;
```

Note that [usernames on GitHub don't allow the inclusion of dots](https://docs.github.com/en/enterprise-cloud@latest/admin/managing-iam/iam-configuration-reference/username-considerations-for-external-authentication#about-username-normalization), so this change seems to be necessary for repository names only.

PR #823, opened 2025-05-19
Update regex github.ts of GitHub Files integration to also capture dots for repo names

boy-vd:
Resolves issue #822 related to GitHub Files integration by updating regexes for capturing URL to a file in a GitHub repository so that it also includes dots for repository names.


addisonschultz (approved):
Nice small change!

PR #860, opened 2025-06-11
Fix LaunchDarkly integration API call to get flags and config screen missing project key input

spastorelli:


integrations/launchdarkly/src/components.tsx
```diff
@@ -86,39 +86,65 @@ export const configBlock = createComponent<
...
-                        <input
-                            label="Secret key"
-                            hint={
-                                <text>
-                                    The secret key from your LaunchDarkly{' '}
-                                    <link
-                                        target={{
-                                            url: 'https://app.LaunchDarkly.co/envs/current/settings/app-environments',
+                    <box>
+                        <vstack>
```
taranvohra, line 90: We need the same `vstack` in `Bucket` too

taranvohra (approved):

integrations/launchdarkly/src/components.tsx
spastorelli: Going to make the change in another PR and merge this for testing.

PR #868, opened 2025-06-12
Update "visitor authentication" to "authenticated access"

talagluck:
Renames visitor authentication to authenticated access. This is primarily meant to be a docs PR, but I ahead and updated the other references within the integration code as well. @addisonschultz please take a look and update or let me know if anything is amiss here.

Also, we'll need to update the names of the visitor authentication example repos:
- https://github.com/GitbookIO/okta-visitor-authentication-example
- https://github.com/GitbookIO/auth0-visitor-authentication-example
- https://github.com/GitbookIO/example-visitor-authentication
- https://github.com/GitbookIO/va-nextjs-clerk
- https://github.com/GitbookIO/visitor-authentication-with-oauth


addisonschultz (approved):
Looks good to me - One small thing is we’ve renamed the visitor auth in the openapi files (in `.gitbook/assets`) but we aren’t using these files anymore anyway - we can merge this PR but let’s also remove them from the space in the GitBook UI after

PR #880, opened 2025-06-23
Add readme for adaptive package

addisonschultz:
This PR adds a readme to the `@gitbook/adaptive` helper package.

### DIFF
--- a/packages/adaptive/README.md
+++ b/packages/adaptive/README.md
@@ -1,8 +1,8 @@
 # `@gitbook/adaptive`
 
-GitBook supports integrations with popular feature flag providers like [LaunchDarkly](https://launchdarkly.com) and [Bucket](https://bucket.co), making it easy to personalize your documentation based on active feature flags.
+GitBook's [adaptive content](https://docs.gitbook.com/help/adaptive-content) feature allows you to personalize your documentation based on visitor data. The `@gitbook/adaptive` SDK provides utilities to help facilitate passing data to GitBook for adapting your content.
 
-This guide walks through how to install and configure each integration using the `@gitbook/adaptive` helper package.
+This SDK includes both feature flag helpers for popular providers like [LaunchDarkly](https://launchdarkly.com) and [Bucket](https://bucket.co), as well as generic utilities for writing custom visitor data.
 
 ---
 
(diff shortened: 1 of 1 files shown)


packages/adaptive/README.md
```diff
@@ -0,0 +1,158 @@
+# `@gitbook/adaptive`
+
+GitBook supports integrations with popular feature flag providers like [LaunchDarkly](https://launchdarkly.com) and [Bucket](https://bucket.co), making it easy to personalize your documentation based on active feature flags.
```
spastorelli: I wonder if we should be more general at the start of the readme when describing what the SDK does as it also exposes generic utils like `writeGitBookVisitorCookie` to allow devs to also write cookies with their own data in case they don't use LD or Bucket. It may also intro additional helpers in the future to be used on the backend side for writing signed cookies (not yet implemented).

Also we should maybe have the first paragraph to introduce adaptive content (maybe link to the docs) and mention that this SDK provide utils to help facilitate passing data to GitBook for adapting their content.

And then add 2 sections:
- first one that talks specifically about FF helper
- another one that talks about the generic one `writeGitBookVisitorCookie`

@taranvohra what do you think?

packages/adaptive/README.md
```diff
@@ -0,0 +1,158 @@
...
+This guide walks through how to install and configure each integration using the `@gitbook/adaptive` helper package.
+
+---
+
+## Prerequisites
+
+Make sure you have:
+
+* A GitBook site with adaptive content enabled
+* An active project on LaunchDarkly or Bucket
```
spastorelli, line 14: I think we should only talk about LD and omit Bucket as we do in the UI:
https://gitbook.slack.com/archives/C08J001JLCD/p1750286067406389

spastorelli (approved):
LGTM on my end. Thanks for doing this 🙏 . Approving to unblock.

Added one comment on changing the intro of the SDK.

packages/adaptive/README.md
addisonschultz: Do you think we should remove Bucket references in general? I think in this readme / docs it’s okay to have - but agree in-product we should focus on launchdarkly. But don’t think we should completely hide information for people who might want to use it

packages/adaptive/README.md
addisonschultz: Added in this feedback

PR #904, opened 2025-07-14
Add scarf integration

addisonschultz:


integrations/scarf/src/scarf-script.raw.js
```diff
@@ -0,0 +1,54 @@
...
+                }
+            }
+            return '';
+        }
+
+        let lastHref = null;
+
+        function sendScarfPing() {
+            const currentHref = window.location.href;
+            if (currentHref === lastHref) return;
```
sdirosa, line 30:
```suggestion
        let lastHref = null;
        
        const cookieGranted = getCookie(GRANTED_COOKIE);

        function sendScarfPing() {
            if(cookieGranted !== 'yes'){
                return;
            }
        
            const currentHref = window.location.href;
            if (currentHref === lastHref) return;
```

And then you can remove the check below because `sendScarfPing` will not send anything unless `cookieGranted` is set to `yes`.

integrations/scarf/src/scarf-script.raw.js
```diff
@@ -0,0 +1,57 @@
...
+            }
+            return '';
+        }
+
+        let lastHref = null;
+
+        const cookieGranted = getCookie(GRANTED_COOKIE);
+
+        function sendScarfPing() {
+            if (cookieGranted !== 'yes') {
```
sdirosa: Hey, sorry about the confusion earlier, I suggested the wrong change.

The correct version should just be:

```js
if (getCookie(GRANTED_COOKIE) !== 'yes') {
```

And there’s no need for the `cookieGranted` variable. 

The reason is: if we read the cookie value once at the top level, it won’t reflect any updates after the user accepts cookies I think. 

The integration would remain disabled even after consent. Reading it dynamically ensures it always checks the latest value.

We need to do the same exact thing in all other integrations - if possible.

sdirosa (approved):
Looks great

PR #907, opened 2025-07-17
Fix missing channels in the configuration of the Slack integration

valentin0h:
This PR aims to address an issue with unpredictable response for the Slack's `conversation.list` API endpoint which results in missing channels from the response.

Splitting the requests by filter appears to reveal some of the missing channels.

### DIFF
--- a/integrations/slack/src/slack.ts
+++ b/integrations/slack/src/slack.ts
@@ -73,8 +73,10 @@ async function fetchChannelsByType(context: SlackRuntimeContext, types: string) 
 export async function getChannelsPaginated(context: SlackRuntimeContext) {
     // Make separate calls for public and private channels due to inconsistent Slack API filtering
     // when requesting multiple types together - some channels are missing from the response
-    const publicChannels = await fetchChannelsByType(context, 'public_channel');
-    const privateChannels = await fetchChannelsByType(context, 'private_channel');
+    const [publicChannels, privateChannels] = await Promise.all([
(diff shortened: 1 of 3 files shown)


integrations/slack/src/slack.ts
```diff
@@ -57,16 +55,34 @@ export async function getChannelsPaginated(context: SlackRuntimeContext) {
...
+ * Executes a Slack API request to fetch channels, handles pagination, then returns the merged
+ * results.
+ */
+export async function getChannelsPaginated(context: SlackRuntimeContext) {
+    // Make separate calls for public and private channels due to inconsistent Slack API filtering
+    // when requesting multiple types together - some channels are missing from the response
+    const [publicChannels, privateChannels] = await Promise.all([
+        fetchChannelsByType(context, 'public_channel'),
+        fetchChannelsByType(context, 'private_channel'),
+    ]);
```
taranvohra, line 79: You need to change the `while` logic to check `subRequests / 2` since you have split the call into 2 separate types

taranvohra (approved):

integrations/slack/src/slack.ts
valentin0h: I doubled the max subrequests. I think the new limit is really 1000 subrequests. These workers don't run on the free tier right? 

https://developers.cloudflare.com/workers/platform/limits/#how-many-subrequests-can-i-make

PR #909, opened 2025-07-18
Log error stacktrace when getting route handling uncaught errors

spastorelli:


gregberge (approved):

integrations/oidc/src/index.tsx
```diff
@@ -382,6 +382,9 @@ const handleFetchEvent: FetchEventCallback<OIDCRuntimeContext> = async (request,
             response = await router.handle(request, context);
         } catch (error: any) {
```
gregberge: Should be unknown, that's unsafe to use any here.

integrations/oidc/src/index.tsx
```diff
@@ -382,6 +382,9 @@ const handleFetchEvent: FetchEventCallback<OIDCRuntimeContext> = async (request,
...
             logger.error('error handling request', error);
+            if ('stack' in error && error.stack) {
```
gregberge: You can do instanceof Error instead

gregberge (approved):

integrations/oidc/src/index.tsx
```diff
@@ -382,6 +382,9 @@ const handleFetchEvent: FetchEventCallback<OIDCRuntimeContext> = async (request,
...
+                logger.error('error stacktrace:', error.stack);
```
emmerich: can you put this into the log above? So you get a single log with everything

integrations/oidc/src/index.tsx
spastorelli: Yes but I think we still need to check existance of stack as it may be inconsitent on the runtime (so just in case):
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/stack

PR #913, opened 2025-07-23
Fix case where oidc integration can redirect to a url ending with double triailling slash

spastorelli:
This can happen in some cases depending on how the auth provide generates the `state` param in the auth response redirection.
And in some cases it can lead to the integration redirect to a URL that has a double slash in it e.g `https://monitoring-release-notes.beyond-suite.com.au//?jwt_token=..`

This has a side effect of GBO or getPublishedContentByURL triggering a redirect to the published site URL with an additional `url/...` in it like:
```
https://monitoring-doc-user.beyond-suite.co/url/monitoring-doc-user.beyond-suite.co
```
[context](https://linear.app/gitbook-x/issue/RND-7704/authenticated-access-redirects-with-a-url-path-in-it#comment-e88ab1b1)

fixes RND-7704

### DIFF
--- a/.changeset/upset-rings-follow.md
+++ b/.changeset/upset-rings-follow.md
@@ -1,0 +1,5 @@
+---
+'@gitbook/integration-oidc': patch
+---
+
+Fix case where oidc integration can redirect to a url ending with double triailling slash


integrations/oidc/src/index.tsx
```diff
@@ -368,9 +368,15 @@ const handleFetchEvent: FetchEventCallback<OIDCRuntimeContext> = async (request,
...
 
                 const state = request.query.state?.toString();
-                const location = state ? state.substring(state.indexOf('-') + 1) : '';
-
-                const url = new URL(`${publishedContentUrl}${location || ''}`);
+                let location = state ? state.substring(state.indexOf('-') + 1) : '';
+                location = location.startsWith('/') ? location.slice(1) : location;
+                const url = new URL(
+                    `${
+                        publishedContentUrl.endsWith('/')
```
spastorelli, line 375: Eventhough `publishedContentUrl` should always include a trailing slash as it comes from the API I prefer to be resilient and do this test just in case.

SamyPesse (approved):

PR #915, opened 2025-07-23
Check if cookies have been accepted before tracking the page

addisonschultz:
This PR should stop integrations from tracking analytics if the user has not accepted cookies or not - One thing I’m not sure, is if the page reloads after the cookie banner is accepted (meaning the cookie is set to yes, and then the integrations will fire?)

@sdirosa can you please have a look and make sure that I’ve implemented this correctly?

### DIFF
--- a/integrations/fathom/src/fathomScript.raw.js
+++ b/integrations/fathom/src/fathomScript.raw.js
@@ -5,9 +5,28 @@
     element.setAttribute('data-site', siteId);
     element.setAttribute('data-spa', 'auto');
     element.src = 'https://cdn.usefathom.com/script.js';
+    const GRANTED_COOKIE = '__gitbook_cookie_granted';
+
+    function getCookie(cname) {
+        const name = `${cname}=`;
+        const decodedCookie = decodeURIComponent(document.cookie);
+        const ca = decodedCookie.split(';');
+        for (let i = 0; i < ca.length; i++) {
+            let c = ca[i];
+            while (c.charAt(0) === ' ') {
+                c = c.substring(1);
+            }
+            if (c.indexOf(name) === 0) {
+                return c.substring(name.length, c.length);
+            }
+        }
+        return '';
+    }
 
     // Function to track external links
     function trackExternalLink(event) {
+        const disableCookies = getCookie(GRANTED_COOKIE) !== 'yes';
+
         var item = event.target.closest('a');
         if (!item) return;
 
@@ -21,6 +40,10 @@
     }
 
     element.onload = function () {
+        if (disableCookies) {
+            return;
+        }
+
         if (trackExternalLinks === true || trackExternalLinks === 'true') {
             // Use event delegation to capture clicks on all current and later loaded links
             doc.addEventListener('click', trackExternalLink);
--- a/integrations/heap/src/script.raw.js
+++ b/integrations/heap/src/script.raw.js
@@ -1,5 +1,29 @@
 (function (win, doc) {
     const trackingID = '<TO_REPLACE>';
+    const GRANTED_COOKIE = '__gitbook_cookie_granted';
+
+    function getCookie(cname) {
+        const name = `${cname}=`;
+        const decodedCookie = decodeURIComponent(document.cookie);
+        const ca = decodedCookie.split(';');
+        for (let i = 0; i < ca.length; i++) {
+            let c = ca[i];
+            while (c.charAt(0) === ' ') {
+                c = c.substring(1);
+            }
+            if (c.indexOf(name) === 0) {
+                return c.substring(name.length, c.length);
+            }
+        }
+        return '';
+    }
+
+    const disableCookies = getCookie(GRANTED_COOKIE) !== 'yes';
+
+    if (disableCookies) {
+        return;
+    }
+
     win.heap = win.heap || [];
     heap.load = function (e, t) {
         (win.heap.appid = e), (win.heap.config = t = t || {});
--- a/integrations/hotjar/src/script.raw.js
+++ b/integrations/hotjar/src/script.raw.js
@@ -1,5 +1,30 @@
 (function (h, o, t, j, a, r) {
     const trackingID = '<TO_REPLACE>';
+
+    const GRANTED_COOKIE = '__gitbook_cookie_granted';
+
+    function getCookie(cname) {
+        const name = `${cname}=`;
+        const decodedCookie = decodeURIComponent(document.cookie);
+        const ca = decodedCookie.split(';');
+        for (let i = 0; i < ca.length; i++) {
+            let c = ca[i];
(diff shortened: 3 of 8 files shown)


integrations/fathom/src/fathomScript.raw.js
```diff
@@ -5,9 +5,28 @@
...
+            let c = ca[i];
+            while (c.charAt(0) === ' ') {
+                c = c.substring(1);
+            }
+            if (c.indexOf(name) === 0) {
+                return c.substring(name.length, c.length);
+            }
+        }
+        return '';
+    }
```
sdirosa, line 24: No clue if it's possible at all, but do we have any way to move this function to a common .js file we may import and reuse in all of these?

integrations/fathom/src/fathomScript.raw.js
```diff
@@ -5,9 +5,28 @@
...
 
     // Function to track external links
     function trackExternalLink(event) {
+        const disableCookies = getCookie(GRANTED_COOKIE) !== 'yes';
```
sdirosa: You have to move this out and on top of this function.

integrations/fathom/src/fathomScript.raw.js
```diff
@@ -21,6 +40,10 @@
     }
 
     element.onload = function () {
+        if (disableCookies) {
+            return;
+        }
```
sdirosa: I'd move this into `trackExternalLink`

integrations/plausible/src/plausibleScript.raw.js
```diff
@@ -6,11 +6,35 @@
...
+        }
+        return '';
+    }
+
     function logIgnoredEvent(event) {
         console.warn('Ignoring Event: ' + event);
     }
 
     function trackEvent(eventName, options) {
+        const disableCookies = getCookie(GRANTED_COOKIE) !== 'yes';
```
sdirosa: Same as the other, you can move this on top and just check it here

integrations/fathom/src/fathomScript.raw.js
addisonschultz: I’m not sure - because each time it’s published it bundles the code I think it’s better to have them duplicate code (also because people might see these as examples, and copy the files in a specific integration so maybe better to have everything encapsulated in one)

integrations/fathom/src/fathomScript.raw.js
```diff
@@ -22,10 +22,13 @@
...
         return '';
     }
+    const disableCookies = getCookie(GRANTED_COOKIE) !== 'yes';
 
     // Function to track external links
     function trackExternalLink(event) {
-        const disableCookies = getCookie(GRANTED_COOKIE) !== 'yes';
+        if (disableCookies) {
+            return;
+        }
```
sdirosa, line 30: @addisonschultz Same here, I suggested the wrong change.

This should be
```js
function trackExternalLink(event) {
      if (getCookie(GRANTED_COOKIE) !== 'yes') {
          return;
      }
...
```

And you can get rid of 
```js
    const disableCookies = getCookie(GRANTED_COOKIE) !== 'yes';
```

integrations/plausible/src/plausibleScript.raw.js
```diff
@@ -29,8 +30,6 @@
     }
 
     function trackEvent(eventName, options) {
-        const disableCookies = getCookie(GRANTED_COOKIE) !== 'yes';
-
         if (disableCookies) {
```
sdirosa: Here you can do:

```js
function trackEvent(event) {
      if (getCookie(GRANTED_COOKIE) !== 'yes') {
          return;
      }
...
```

and `disableCookies` can go.

integrations/heap/src/script.raw.js
```diff
@@ -1,5 +1,29 @@
...
+            }
+        }
+        return '';
+    }
+
+    const disableCookies = getCookie(GRANTED_COOKIE) !== 'yes';
+
+    if (disableCookies) {
+        return;
+    }
```
sdirosa, line 23: This approach will not re-enable the integration when the cookie changes. Unless the page/script gets completely reloaded.

Do we do that after a user accepts the banner?

integrations/hotjar/src/script.raw.js
```diff
@@ -1,5 +1,29 @@
...
+            }
+        }
+        return '';
+    }
+
+    const disableCookies = getCookie(GRANTED_COOKIE) !== 'yes';
+
+    if (disableCookies) {
+        return;
+    }
```
sdirosa, line 23: Same here, if we don't reload page/script on cookie accept, this won't work.

integrations/mixpanel/src/script.raw.js
```diff
@@ -1,4 +1,28 @@
...
+            }
+        }
+        return '';
+    }
+
+    const disableCookies = getCookie(GRANTED_COOKIE) !== 'yes';
+
+    if (disableCookies) {
+        return;
+    }
```
sdirosa, line 22: Same here

integrations/posthog/src/script.raw.js
```diff
@@ -1,7 +1,29 @@
...
+        }
+    }
+    return '';
+}
+const disableCookies = getCookie(GRANTED_COOKIE) !== 'yes';
 
 !(function (t, e) {
+    if (disableCookies) {
+        return;
+    }
```
sdirosa, line 24: Same. Without reloading this won't reactivate. The script below is a minified mess, so I'm not even sure what it does 🤣 

I'll test a bit tomorrow and see if there's a way to achieve what we want here. There may be a way to preconfigure the posthog SDK without preventing its initialization.

integrations/reo/src/script.raw.js
```diff
@@ -1,6 +1,29 @@
...
+    }
+    return '';
+}
+
+const disableCookies = getCookie(GRANTED_COOKIE) !== 'yes';
 
 (function (r, e, o) {
+    if (disableCookies) {
+        return;
+    }
```
sdirosa, line 23: Same for this one. It looks like their SDK injects a script in the page. If we prevent that and re-enabling the cookie won't rerun this function, then it's never get initialized again (unless I'm mistaken). 

So we should check if there's a way to initialize reo in disabled mode and then programmatically re-activate it when the cookie changes. Or skip the tracking.

integrations/heap/src/script.raw.js
addisonschultz: Yes, it looks like the page is reloaded after the banner is accepted/rejected

sdirosa:
> I’ve moved the check for the cookies into the function itself instead of a variable. The page should reload when the cookies are accepted/rejected anyway, though

If that's the case then we are good 👌🏼

PR #932, opened 2025-08-11
HubSpot conversations

valentin0h:
- [x] Create HubSpot staging app
  - [ ]  Create HubSpot production app


integrations/hubspot-conversations/src/config.tsx
```diff
@@ -0,0 +1,44 @@
...
+                                    ? 'Re-authorize'
+                                    : 'Authorize'
+                            }
+                            onPress={{
+                                action: '@ui.url.open',
+                                url: `${installation?.urls.publicEndpoint}/oauth`,
+                            }}
+                        />
+                    }
+                />
```
valentin0h, line 40: Add minimal message showing that the webhooks are active, otherwise there's just a auth button.

integrations/hubspot-conversations/src/client.ts
```diff
@@ -0,0 +1,127 @@
...
+        throw new Error('HubSpot OAuth credentials not found');
+    }
+
+    return await getOAuthToken(oauth_credentials, getHubSpotOAuthConfig(context), context);
+}
+
+/**
+ * Make a request to the HubSpot API.
+ */
+export async function hubspotApiRequest<T = unknown>(
```
SamyPesse, line 94: Maybe you can use https://github.com/HubSpot/hubspot-api-nodejs
to get better typing

integrations/hubspot-conversations/src/conversations.ts
```diff
@@ -0,0 +1,156 @@
...
+        const conversations = response.results || [];
+        logger.info('Found conversations', { count: conversations.length, page: pageIndex });
+
+        if (conversations.length === 0) {
+            break;
+        }
+
+        // Filter for closed conversations only
+        const closedConversations = conversations.filter(
+            (conv) => conv.latestMessageTimestamp && conv.status === 'CLOSED',
```
SamyPesse: Is it possible to filter this in the API call?

integrations/hubspot-conversations/src/index.ts
```diff
@@ -0,0 +1,70 @@
...
+                context.environment.spaceInstallation?.urls?.publicEndpoint ||
+                    context.environment.installation?.urls.publicEndpoint ||
+                    context.environment.integration.urls.publicEndpoint,
+            ).pathname,
+        });
+
+        /*
+         * Webhook to ingest conversations when they are closed.
+         */
+        router.post('/webhook', async (request) => {
```
SamyPesse, line 24: How is the webhook configured? Globally or per installation?

I don't see anything about it

SamyPesse (approved):
Nice! A few comments, but nothing blocking as it only publishes in staging for now and we can test there :)

integrations/hubspot-conversations/src/index.ts
valentin0h: We configure it globally, once, in the hubspot app . E.g `https://<your-integration-env>/v1/integrations/hubspot-conversations/integration/webhook`. 

When the app is installed in a hubspot account the webhook is active for that installation. We can match the hubspot installation with gitbook installations via `externalIds`, like we usually do.

integrations/hubspot-conversations/src/client.ts
valentin0h: It doesn't seem to support the endpoints we care about.

PR #940, opened 2025-08-13
Various fixes for intercom-conversations

valentin0h:
## Fixes
- Fix webhook event (`conversation.closed` → `conversation.admin.closed`)
- Fix matching of intercom workspace with gitbook installation via `externalId`
- Better batching of ingested conversations
- Fix search query (`open: false` → `state: 'closed'`)

## Improvements
• Multi-installation support for webhooks
• More cautious with workers subrequest limits (700 vs 1000+ requests)

## Follow-up
- [ ] Webhook signature verification will be implemented in next PR


integrations/intercom-conversations/src/client.ts
```diff
@@ -10,9 +12,66 @@ export function getIntercomOAuthConfig(context: IntercomRuntimeContext) {
...
+                );
+            }
+
+            logger.debug('Intercom OAuth response received', {
+                hasAccessToken: !!response.access_token,
+                tokenLength: response.access_token?.length,
+            });
+
+            // Get workspace information using the access token
+            // We use fetch here instead of IntercomClient because this is during OAuth setup
```
SamyPesse: I'm not sure to understand why we can't just use it here, you could pass the toke n to it

integrations/intercom-conversations/src/client.ts
```diff
@@ -25,12 +84,12 @@ export async function getIntercomClient(context: IntercomRuntimeContext) {
     const { installation } = context.environment;
 
     if (!installation) {
-        throw new Error('Installation not found');
+        throw new ExposableError('Installation not found');
```
SamyPesse: Why should it be exposable error? it should not happen

integrations/intercom-conversations/src/conversations.ts
```diff
@@ -1,34 +1,37 @@
...
+        throw new ExposableError('Installation not found');
     }
 
-    const intercom = await getIntercomClient(context);
+    const intercomClient = await getIntercomClient(context);
 
     let pageIndex = 0;
     const perPage = 100;
-    const maxPages = 10;
+    const maxPages = 7; // Keep under ~1000 subrequest limit. Calc: 7 pages * 100 items ≈ 700 detail calls + 7 search page calls ≈ ~707 Intercom calls (+7 GitBook ingests ≈ ~714 total).
```
SamyPesse, line 23: Makes sense 👍

SamyPesse (approved):

integrations/intercom-conversations/src/client.ts
valentin0h: Wrong comment; It's not exposed in `intercom-client`

PR #941, opened 2025-08-14
Add cookie banner to tracking integraitons

addisonschultz:
Should fix https://github.com/orgs/GitbookIO/discussions/1055

As it stands, the cookie banners do not show because we were not showing them through the `site:script:cookies` scope. 

We recently added a change to the integrations to not fire analytics links if the cookie banner was not accepted, causing the bug in the community bug above. 

This should fix all affected integrations


sdirosa (approved):
👌🏼

PR #944, opened 2025-08-15
Adds a no-cache header to slack error responses

valentin0h:
Made the wrong fix https://github.com/GitbookIO/integrations/pull/935. The error response is what gets cached first due to the lack of authentication permissions.

I want to test if this works as expected before I make any changes to gitbook-x

### DIFF
--- a/integrations/slack/src/router.ts
+++ b/integrations/slack/src/router.ts
@@ -1,6 +1,11 @@
 import { Router } from 'itty-router';
 
-import { createOAuthHandler, FetchEventCallback, OAuthResponse } from '@gitbook/runtime';
+import {
+    createOAuthHandler,
+    ExposableError,
+    FetchEventCallback,
+    OAuthResponse,
+} from '@gitbook/runtime';
 
 import {
     createSlackEventsHandler,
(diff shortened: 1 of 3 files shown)


integrations/slack/src/router.ts
```diff
@@ -119,19 +124,34 @@ export const handleFetchEvent: FetchEventCallback = async (request, context) =>
...
 
-        return new Response(JSON.stringify(completions), {
-            headers: {
-                'Content-Type': 'application/json',
-                'Cache-Control': 'no-cache, no-store',
-            },
-        });
+            return new Response(JSON.stringify(completions), {
+                headers: {
+                    'Content-Type': 'application/json',
```
valentin0h, line 137: removed caching from here, let gitbook handle caching

integrations/slack/src/router.ts
```diff
@@ -119,19 +124,34 @@ export const handleFetchEvent: FetchEventCallback = async (request, context) =>
...
+            return new Response(
+                JSON.stringify({
+                    error:
+                        error instanceof ExposableError ? error.message : 'Internal server error',
+                }),
+                {
+                    status: error instanceof ExposableError ? error.code : 500,
+                    headers: {
+                        'Content-Type': 'application/json',
+                        'Cache-Control': 'no-cache, no-store',
```
valentin0h, line 150: Error responses shouldn't be cached.

valentin0h: The idea is to fix the logic in gitbook, not to cache error responses, but I want to validate that this works regardless.

valentin0h:
Fixed in gitbook

PR #947, opened 2025-08-18
Add native AI experience for RunLLM in GitBook

addisonschultz:
Hi @likawind 

This PR adds extra functionality to the RunLLM integration to tap into GitBook native UI.

### What we’ve added to GitBook
We’ve introduced a new feature to allow AI integrations to use the native GitBook UI/search, instead of defaulting to the GitBook assistant. This will allow GitBook users to choose what type of AI experience they want. Previously, this was not possible, meaning many GitBook sites using AI integrations would have 2 experiences - the GitBook Assistant and a floating button/AI, such as RunLLM.

#### UI/Video

https://github.com/user-attachments/assets/39e1bd31-f787-44c6-b13b-fed1019a5b53

Note: This video shows the GitBook Assistant still enabled, causing there to be 2 buttons for AI chat in the UI (next to the search bar). Ideally, users would disable the GitBook Assistant after adding this integration, meaning only one AI button would appear.

### What this PR does
This PR adds the newly created API from the integration platform to allow AI integrations to tap into the GitBook UI. It also introduces an extra configuration native_ai_experience to allow users to select the AI experience they want (Native using GitBook UI, or Non-native (continue using the floating button in the lower corner).

The call for `registerCustomAssistant` has a few properties that can be changed.

Required props:
`label`: The name of the AI assistant
`icon`: An icon to display in the UI to represent the assistant. Must be an icon from https://fontawesome.com/. Custom icons are not supported at this time.
`open`: Function called when the UI button (next to the search bar) is pressed or question is asked. Meant to open the RunLLM modal, in this instance. Contains an argument query which contains a question when asked from the search UI. 

Image for reference:
image

Optional prop:
`ui?`: Optional property whether to display the UI for this integration. This should be used if a user does not want to show the button in the UI (next to the search bar), but would like to continue to use the floating button in the lower right (injected from RunLLM). The reason we introduced this flag, is because we still want integrations to register through this new API, allowing users to tap into the search/ask workflow to ask questions and still fire the RunLLM modal. (See image above)

### What still needs to be done/discussed
1. In the event the user wants to use the Native experience (Using the GitBook button to fire the RunLLM modal), we want to disable/hide the floating button from RunLLM. Is there a reliable query selector we can use to set it’s visibility to none?
2. The open function is not finished yet, as I need to discuss with you on what method could be used to open the RunLLM modal. Is there a function or call exposed by the script that is available for this integration to call?


.changeset/fancy-states-laugh.md
SamyPesse, line 1: You don't need to add a changeset for every commit 😅 
A changeset should represent the entire chage made by this PR

SamyPesse: You can delete these files if you want and only keep one

integrations/runllm-widget/src/script.raw.js
```diff
@@ -45,4 +46,28 @@
...
+    } else {
+        window.GitBook.registerAssistant({
+            label: 'RunLLM',
+            icon: 'sparkle',
+            ui: false,
+            open: (query) => {
+                // Open RunLLM window with query
+            },
+        });
+    }
```
SamyPesse: It doesn't look finished;

also you can simplify as:

```js
if (nativeAiExperience) {
  // Hide the widget
}

window.GitBook.registerAssistant({
            label: 'RunLLM',
            icon: 'sparkle',
            ui: nativeAiExperience,
            open: (query) => {
                // Open RunLLM window with query
            },
});
```

integrations/runllm-widget/src/script.raw.js
```diff
@@ -9,6 +9,7 @@
     const communityUrl = '<COMMUNITY_URL>';
     const communityType = '<COMMUNITY_TYPE>';
     const disableAskAPerson = '<DISABLE_ASK_A_PERSON>';
+    const nativeAiExperience = '<NATIVE_AI_EXPERIENCE>';
```
SamyPesse:
```suggestion
    const nativeAiExperience = <NATIVE_AI_EXPERIENCE>;
```

So that you can use it as a boolean

integrations/runllm-widget/src/script.raw.js
addisonschultz: Yes not finished yet - Having a call with them soon to finish this - thanks for the review though 👍

integrations/runllm-widget/src/script.raw.js
addisonschultz: I don’t think this works? Getting a linting error - this is a .js file and it’s reading as a tag

integrations/runllm-widget/gitbook-manifest.yaml
```diff
@@ -46,6 +46,11 @@ contentSecurityPolicy:
 configurations:
     site:
         properties:
+            native_ai_experience:
+                type: boolean
+                title: Native AI experience
+                description: Set to "true" to enable the built-in AI experience. Default is "false".
```
likawind: Should we mention that this will hide the RunLLM button? For example:

"Set to "true" to enable triggering RunLLM from githbook's built-in AI experience. This will hide the RunLLM trigger button. Default is "false"."

likawind (approved):
Exciting to see this @addisonschultz , thank you for getting us integrated with gitbook more smoothly

PR #948, opened 2025-08-19
Fix segment tracking

addisonschultz:
This PR amends tracking labels in Segment integration. Also creating PR to make sure integration is published properly

### DIFF
--- a/integrations/segment/src/events.ts
+++ b/integrations/segment/src/events.ts
@@ -6,12 +6,12 @@ import { version } from '../package.json';
  * Generate the event to track in Segment for an actual GitBook event.
  */
 export function generateSegmentTrackEvent(event: api.SiteViewEvent) {
-    const { visitor, referrer, url, spaceId, pageId, siteId } = event;
+    const { visitor, referrer, url, siteId } = event;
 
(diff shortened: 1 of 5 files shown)


integrations/segment/src/events.ts
```diff
@@ -6,12 +6,12 @@ import { version } from '../package.json';
...
  */
 export function generateSegmentTrackEvent(event: api.SiteViewEvent) {
-    const { visitor, referrer, url, spaceId, pageId, siteId } = event;
+    const { visitor, referrer, url, siteId } = event;
 
     const anonymousId = getAnonymousId(event);
     const visitedURL = new URL(url);
     return {
-        event: '[GitBook] space_view',
+        event: '[GitBook] site_view',
```
taranvohra, line 14: Why do we need `[GitBook]`?

taranvohra (approved):

integrations/segment/src/events.ts
addisonschultz: It will show up in the segment side of things - might be nice to have in case users have multiple sources so they know which ones are coming from GitBook

PR #954, opened 2025-08-20
Add configuration to fix tracking data in EU

### LINKED ISSUE #950: Segment integration doesn't work for EU workspaces
## Issue

As stated in the [Segment HTTP API documentation](https://segment.com/docs/connections/sources/catalog/libraries/server/http-api/):

> HTTP API sources in EU workspaces should use the `events.eu1.segmentapis.com` endpoint

> If you are located in the EU and use the https://api.segment.io/v1/ endpoint, you might not see any errors, but your events will not appear in the Segment app.

The Segment URL is hard-coded here:
https://github.com/GitbookIO/integrations/blob/main/integrations/segment/src/index.ts#L25

Meaning no data is received in our workspace.

Was able to confirm on our Segment HTTP API Source.

This returns `"success": true` but nothing shows up in our workspace:
```
curl --location 'https://api.segment.io/v1/track' \
  --header 'Content-Type: application/json' \
  --data-raw '{
      "event": "happy-path-a3ef8a6f-0482-4694-bc4d-4afba03a0eab",
      "email": "test@example.org",
      "userId": "123",
      "writeKey": "xxxxxxxxxxxxxxxxxxxxxxxxx"
  }'
```

This works correctly:
```
curl --location 'https://events.eu1.segmentapis.com/v1/track' \
  --header 'Content-Type: application/json' \
  --data-raw '{
      "event": "happy-path-a3ef8a6f-0482-4694-bc4d-4afba03a0eab",
      "email": "test@example.org",
      "userId": "123",
      "writeKey": "xxxxxxxxxxxxxxxxxxxxxxxxx"
  }'
```

## Suggestion
Allow the user to choose the Segment hostname from a dropdown containing `api.segment.io` and `events.eu1.segmentapis.com`.

Document this on the integration page.

addisonschultz:
Fixes #950 

We are using the wrong endpoint for customers located in the EU, meaning events are not successfully tracked. 

This PR adds a configuration to select if you are in the EU to use a different endpoint.

More info in the segment docs: https://segment.com/docs/connections/sources/catalog/libraries/server/http-api/

### DIFF
--- a/integrations/segment/gitbook-manifest.yaml
+++ b/integrations/segment/gitbook-manifest.yaml
@@ -31,7 +31,7 @@ summary: |
 
     In Segment, you can set up a source to receive the GitBook events. After adding an HTTP API source in Segment, you will see a Write Key. This key is used to identify the data source in GitBook.
 
-    If you are located in the EU, you may need to configure an optional parameter to track events successfully.
+    If you are [**located in the EU**](https://segment.com/docs/guides/eu-region/), you may need to configure an optional parameter to track events successfully.
 
     To install the Segment integration on a single space click on the Integrations button in sub-navigation. Alternatively, you can install it on multiple or all spaces by going into your organization settings. You will need the Segment Write Key to help you identify the data source.
 categories:
@@ -42,7 +42,7 @@ configurations:
             write_key:
                 type: string
                 title: Write Key
-                description: Key provided by Segment to identify the data source.
+                description: Key provided by Segment to identify the data source. If you are located in the EU, you will need to use an EU Write Key.
             eu_region:
                 type: boolean
                 title: EU Region
--- a/integrations/segment/gitbook-manifest.yaml
+++ b/integrations/segment/gitbook-manifest.yaml
@@ -31,7 +31,7 @@ summary: |
 
     In Segment, you can set up a source to receive the GitBook events. After adding an HTTP API source in Segment, you will see a Write Key. This key is used to identify the data source in GitBook.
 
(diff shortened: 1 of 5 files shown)


integrations/segment/gitbook-manifest.yaml
```diff
@@ -31,6 +31,8 @@ summary: |
 
     In Segment, you can set up a source to receive the GitBook events. After adding an HTTP API source in Segment, you will see a Write Key. This key is used to identify the data source in GitBook.
 
+    If you are located in the EU, you may need to configure an optional parameter to track events successfully.
```
spastorelli: I would maybe bold `**located in the EU**` and maybe add a link to https://segment.com/docs/guides/regional-segment/ for more info.

Also should we maybe mention that the write Key needs to be a EU region writeKey when they use this option since:
> Data sent to the EU endpoint using a US-region write key will get rejected.

https://segment.com/docs/guides/regional-segment/#custom-http-requests

integrations/segment/gitbook-manifest.yaml
```diff
@@ -41,6 +43,11 @@ configurations:
                 type: string
                 title: Write Key
                 description: Key provided by Segment to identify the data source.
+            eu_region:
+                type: boolean
```
spastorelli, line 47: nit: I wonder if **in the future** we should have a Segment region select input field instead with US (default) and EU as options as it seems they may expand regional segments beyond EU in the future:
> Regional Segment is currently limited to the EU. Future expansion of Regional Segment beyond the EU is under evaluation by Segment Product and R&D.

https://segment.com/docs/guides/regional-segment/#known-limitations

Not something to do right now but to keep in mind in the future.

spastorelli (approved):

PR #965, opened 2025-08-29
Improve slack notifications

valentin0h:
## Summary
• Enhanced Slack notification format with modern block-based UI instead of plain text
• Added structured sections for different types of changes (new/modified/deleted pages and files)
• Included action buttons for easy access to main space and published docs
• Added emoji indicators and better visual hierarchy for improved readability

**before**
CleanShot 2025-08-29 at 14 13 12@2x

**after**
CleanShot 2025-08-29 at 14 13 59@2x

### DIFF
--- a/integrations/slack/src/types.ts
+++ b/integrations/slack/src/types.ts
@@ -1,23 +1,23 @@
 // Slack Block Kit Types
-export type TTextType = 'plain_text' | 'mrkdwn';
+export type SlackTextType = 'plain_text' | 'mrkdwn';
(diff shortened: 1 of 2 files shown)


integrations/slack/src/types.ts
taranvohra, line 1: Why prefix with `T` instead of `Slack`? like `SlackBlock`

valentin0h: 🤖 🙈

PR #973, opened 2025-09-05
Update `@gitbook/adaptive` SDK to add a withReflag helper and deprecate withBucket one

spastorelli:


valentin0h (approved):

packages/adaptive/src/reflag.ts
```diff
@@ -46,3 +49,49 @@ export function withBucket(client: BucketClient): () => void {
         cleanup.forEach((fn) => fn());
     };
 }
+
+/**
+ * Sets a client-side session cookie with the features enabled in the reflag client.
+ */
+export function withReflag(client: ReflagClient): () => void {
```
gregberge, line 56: Why is the whole content duplicated? Why not just creating one abstraction?

packages/adaptive/src/reflag.ts
```diff
@@ -1,8 +1,11 @@
 import type { BucketClient } from '@bucketco/browser-sdk';
```
gregberge, line 1: Why still using this client? It's just a rename so just use the new package as it was an update?

spastorelli: The types are different. So until we sunset the withBucket helper we need to use both types.

packages/adaptive/src/reflag.ts
spastorelli: The withBucket is deprecated and we will be sunsetting it. I didn't think it was necessary to create an abstraction.

gregberge: Hmm OK, if it's tree shaked yes.

packages/adaptive/src/reflag.ts
gregberge: OK, why do you want to keep the bucket one? Why not just moving it to reflag and that's it?

spastorelli: Because we need to keep the withBucket helper for customers who still use it until they upgrade to new version of our gitbook/adaptive package. Btw we don't actually depend on the bucket/reflag client as client instance is "dep-injected" to the helper. Our dep on the bucket/reflag SDK is just a dev dep for the type.

gregberge: Just release a major of `@gitbook/adaptive` and remove `withBucket`, I think that's fine.

spastorelli: https://github.com/GitbookIO/integrations/pull/979

PR #989, opened 2025-09-16
Webhook integration

sdirosa:
Webhook Integration. 

Docs: https://github.com/GitbookIO/integrations/blob/surai/rnd-7453-webhook-integration/docs/guides/webhook.md

### Overview
CleanShot 2025-09-19 at 09 59 36

### Configuration screen
CleanShot 2025-09-19 at 10 00 34

### Example of a received event
CleanShot 2025-09-18 at 19 52 40@2x


integrations/webhook/src/index.ts
```diff
@@ -0,0 +1,153 @@
...
+    // Add HMAC signature for webhook verification
+    const encoder = new TextEncoder();
+    const key = await crypto.subtle.importKey(
+        'raw',
+        encoder.encode(config.secret),
+        { name: 'HMAC', hash: 'SHA-256' },
+        false,
+        ['sign'],
+    );
+    const signatureBuffer = await crypto.subtle.sign('HMAC', key, encoder.encode(jsonPayload));
```
spastorelli: Would it be worth generating the hash based also on a timestamp, at which we send the event, to prevent against replay attacks? We could use similar format as Stripe uses e.g: `<timestamp>.<JSONPayload>` for the payload that we hash and include the timestamp in the signature header.

I know that customers should make sure that their webhook handlers are idempotent to prevent replay attacks but generating timestamp based signature would enforce protection if they don't ensure idempotency.

integrations/webhook/src/config.tsx
```diff
@@ -0,0 +1,140 @@
...
+                        label={AVAILABLE_EVENTS[eventType]}
+                        element={<switch state={eventType} />}
+                    />
+                ))}
+
+                <divider size="medium" />
+
+                <markdown content="### Secret" />
+                <text>
+                    You can copy and verify this secret in your webhook for HMAC verification:
```
spastorelli:
```suggestion
                    Use this secret to verify the signature included in GitBook webhook events (i.e HMAC with SHA-256 verification):
```

We can maybe at some stage link to a section in our docs that talks about webhooks (how to verify signature and example of events objects)

integrations/webhook/src/index.ts
```diff
@@ -0,0 +1,153 @@
...
+                    status: response.status,
+                    statusText: response.statusText,
+                });
+
+                // Retry on server errors (5xx) and rate limiting (429)
+                if (
+                    retryCount < MAX_RETRIES &&
+                    (response.status >= 500 || response.status === 429)
+                ) {
+                    return retryWithDelay(
```
spastorelli: How long could the integration script be delayed in the worst-case retry scenario? I’m not sure what the upper limit is for Cloudflare Workers.

Not sure if it's worth but we could consider using the `api.integrations.queueIntegrationTask` API (similar to how the GitHub integration works) to handle this.

But if the whole retry window is reasonable then probably no need.

spastorelli (approved):
Nice 👌

integrations/webhook/src/index.ts
sdirosa: Good question!

So for the exponential backoff, counting also the worst possible jitter, it should be:
1000+100 + 2000+200 + 4000+400 = 7700 ms

But, if there's a weird connectivity issue we could hit the request timeouts and then it becomes:
10000+1100 + 10000+2200 + 10000+4400  for a total of almost 38 seconds 

I think workers should be totally fine with these waits

integrations/webhook/src/config.tsx
sdirosa: Yes, good idea 💯 
Most of this stuff is fairly standard, but giving some examples it's definitely a lot nicer. I'll add some docs.

integrations/webhook/src/index.ts
sdirosa: That's not a bad idea, I'll give it a shot 👌🏼

sdirosa: Updated as suggested

integrations/webhook/src/index.ts
sdirosa: Actually, I just checked and I think this won't work exactly as intended.

From the docs:

> Due to [security-based restrictions on timers](https://developers.cloudflare.com/workers/reference/security-model/#step-1-disallow-timers-and-multi-threading) in Workers, timers are limited to returning the time of the last I/O. This means that while setTimeout, setInterval, and setImmediate will defer your function execution until after other events have run, they will not delay them for the full time specified.

sdirosa: I'm going to look into `api.integrations.queueIntegrationTask` and see how that would work

sdirosa: I implemented `queueIntegrationTask` for retries and added a bunch of tests. I'm going to build a stupid server to test it properly tomorrow manually as well.

EDIT. Actually I tested with the normal retry strategy I had and it works fine, I tried simulating a webhook failing 2 times and succeeding on the third and the delays were all correct.

I think we are good and it's a lot simpler

integrations/webhook/src/config.tsx
sdirosa: I added a dedicated docs page for the webhook

PR #994, opened 2025-09-19
Webhook integration - 2nd pass

sdirosa:
Second iteration on the webhook integration.

Adds:
- Dynamic configuration panels for the integration, to show only events relevant to an installation
- Store secret and webhook at org level, so they get inherited in all installations
- Fix site events
- Improve docs to clarify site vs spaces events

### Site config
CleanShot 2025-09-19 at 14 45 09

### Space config
CleanShot 2025-09-19 at 14 46 26

### DIFF
--- a/.changeset/smart-memes-lick.md
+++ b/.changeset/smart-memes-lick.md
@@ -1,0 +1,5 @@
+---
+'@gitbook/integration-webhook': minor
+---
+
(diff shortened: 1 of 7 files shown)


sdirosa:
@spastorelli Since you have context already, could you check these improvements and fixes?
Thanks!

spastorelli (approved):

PR #995, opened 2025-09-19
Fix webhook integration docs links

sdirosa:
They were moved, but the manifest and config were still pointing to the old url. 
Took the opportunity to use the link from the manifest in the config as well.


integrations/webhook/src/config.tsx
sdirosa, line 1: In short, here we merged the configs into one panel. Webhook url and secret are account level variables and always loaded and saved from there - thus shared across all installs. The events are dynamic and vary depending whether we are in a site or space and they are always stored at the installation level. Single button to save both. 

In the future, if there's feedback about it, we could make webhook url and secret overridable at site/space installation level, but for now I think this is much simpler and easier to use.

spastorelli (approved):

PR #1011, opened 2025-09-25
Use integration tasks for processing Ask AI queries in Slack

taranvohra:
When we receive a slack event for an Ask AI query it can sometimes take more than 30s for the AI API to respond. 30s is the timeout limit of [`waitUntil` in cf workers ](https://developers.cloudflare.com/workers/platform/limits/#duration)

This PR changes the strategy where we queue an async task for the AI query so it's not subject to the 30s limit of waitUntil

### DIFF
--- a/integrations/slack/src/router.ts
+++ b/integrations/slack/src/router.ts
@@ -6,7 +6,7 @@ import {
     FetchEventCallback,
     Logger,
     OAuthResponse,
-    verifyIntegrationSignature,
+    verifyIntegrationRequestSignature,
 } from '@gitbook/runtime';
 
 import {
@@ -60,7 +60,7 @@ export const handleFetchEvent: FetchEventCallback = async (request, context) => 
      * Handle integration tasks
      */
     router.post('/tasks', async (request) => {
-        const verified = await verifyIntegrationSignature(request, environment);
+        const verified = await verifyIntegrationRequestSignature(request, environment);
 
         if (!verified) {
             const message = `Invalid signature for integration task`;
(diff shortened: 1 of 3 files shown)


integrations/slack/src/actions/queryAskAI.ts
```diff
@@ -149,59 +154,127 @@ async function getRelatedSources(params: {
...
+    params: IQueryAskAI & {
+        query: string;
+        installation: IntegrationInstallation;
+        accessToken: string | undefined;
+    },
+) {
+    const { accessToken, installation, query, context, ...rest } = params;
+
+    // Authenticate as the integration
+    const gitbook = new GitBookAPI({
```
spastorelli: Can't you get the client passed from the caller that could get it from the context?

integrations/slack/src/router.ts
```diff
@@ -45,6 +56,67 @@ export const handleFetchEvent: FetchEventCallback = async (request, context) =>
         ].join(' '),
     );
 
+    async function verifyIntegrationSignature(
```
spastorelli: Not for this PR but I think this is dupped across multiple integrations. It would be nice to have something factored out (along with some tests) into a shared package (maybe the runtime).

integrations/slack/src/utils.ts
```diff
@@ -106,8 +117,34 @@ export function getActionNameAndType(actionId: string) {
...
+ * Convert an array buffer to a hex string
+ */
+export function arrayToHex(arr: ArrayBuffer) {
+    return [...new Uint8Array(arr)].map((x) => x.toString(16).padStart(2, '0')).join('');
+}
+
+/**
+ * Constant-time string comparison. Equivalent of `crypto.timingSafeEqual`.
+ **/
+export function safeCompare(expected: string, actual: string) {
```
spastorelli: Same for this one.

spastorelli (approved):

PR #1017, opened 2025-10-07
Salesforce chat integration

sdirosa:
Salesforce embedded service chat integration.

For reference, here is the original snippet:

```js
<script type='text/javascript'>
	function initEmbeddedMessaging() {
		try {
			embeddedservice_bootstrap.settings.language = 'en_US';

			embeddedservice_bootstrap.init(
				'<ORG_ID>',
				'<MESSAGING_DEPLOYMENT_NAME>',
				'<EMBEDDED_SERVICE_SITE_URL>',
				{
					scrt2URL: '<SCRT2_URL>'
				}
			);
		} catch (err) {
			console.error('Error loading Embedded Messaging: ', err);
		}
	};
</script>
<script type='text/javascript' src='<EMBEDDED_SERVICE_SITE_URL>/assets/js/bootstrap.min.js' onload='initEmbeddedMessaging()'></script>

```

### Preview
CleanShot 2025-10-08 at 19 53 38

Implements [RND-7171](https://linear.app/gitbook-x/issue/RND-7171/servicecloud-integration)

### DIFF
--- a/integrations/salesforce-chat/gitbook-manifest.yaml
+++ b/integrations/salesforce-chat/gitbook-manifest.yaml
@@ -17,19 +17,24 @@ contentSecurityPolicy:
     script-src: |
         https://service.force.com
         https://*.force.com
-        https://*.salesforce.com;
+        https://*.salesforce.com
+        https://*.salesforce-scrt.com;
     connect-src: |
         https://*.force.com
         https://*.salesforce.com
-        wss://*.salesforce.com;
+        https://*.salesforce-scrt.com
+        wss://*.salesforce.com
+        wss://*.salesforce-scrt.com;
     img-src: |
         data:
         blob:
         https://*.force.com
-        https://*.salesforce.com;
+        https://*.salesforce.com
+        https://*.salesforce-scrt.com;
     frame-src: |
         https://*.force.com
-        https://*.salesforce.com;
+        https://*.salesforce.com
+        https://*.salesforce-scrt.com;
     style-src: |
         'unsafe-inline';
 summary: |


integrations/salesforce-chat/src/script.raw.js
```diff
@@ -0,0 +1,71 @@
...
+                ORG_ID,
+                MESSAGING_DEPLOYMENT_NAME,
+                EMBEDDED_SERVICE_SITE_URL,
+                {
+                    scrt2URL: SCRT2_URL,
+                },
+            );
+        } catch (err) {
+            // initialization failed
+        }
```
sdirosa, line 38: Looking at other integrations, seems like we never log nor report these kind of errors, so I left it like this for now

integrations/salesforce-chat/package.json
```diff
@@ -0,0 +1,19 @@
+{
+    "name": "@gitbook/integration-salesforce-chat",
```
sdirosa, line 2: Not sure if we should call this just `salesforce`, like the `intercom` one.

integrations/salesforce-chat/gitbook-manifest.yaml
```diff
@@ -0,0 +1,63 @@
...
+    # Overview
+
+    This integration injects Salesforce Service Cloud Chat into your published GitBook site, enabling a Salesforce support chat for your users.
+
+    # Configure
+
+    Provide the required Salesforce identifiers and URLs from your Embedded Service deployment in Salesforce Setup.
+categories:
+    - analytics
+configurations:
```
sdirosa, line 50: - Needs assets
- Check summary copy
- Not sure about the category
- We probably need a documentation link

integrations/salesforce-chat/package.json
addisonschultz: I would call this salesfore-chat (Salesforce Chat)? I think better than just calling it Salesforce as that might be too broad for an integration name for Salesforce

integrations/salesforce-chat/gitbook-manifest.yaml
addisonschultz: Updated

integrations/salesforce-chat/gitbook-manifest.yaml
```diff
@@ -0,0 +1,69 @@
+name: salesforce-chat
+title: Salesforce Service Cloud Chat
+organization: gitbook
+description: Add Salesforce Service Cloud Chat to your published GitBook content
+visibility: public
+script: ./src/index.ts
+icon: ./assets/salesforce-icon.png
+previewImages:
+    - ./assets/salesforce-visual.png
```
sdirosa, line 9: @addisonschultz I fixed the urls for the icon and preview image, could you double check?

PR #1030, opened 2025-10-30
Add support for ingesting conversation in Slack integration

spastorelli:
This PR aims to add support for ingesting conversations via a command & shortcut in the Slack integration.

It does **not** yet add support for app mentions yet as this will be done in another PR.

Fixes RND-8529

### DIFF
--- a/integrations/slack/src/actions/ingestConversation.ts
+++ b/integrations/slack/src/actions/ingestConversation.ts
@@ -1,7 +1,11 @@
 import { ConversationInput, ConversationPart, IntegrationInstallation } from '@gitbook/api';
 import { SlackInstallationConfiguration, SlackRuntimeContext } from '../configuration';
 import { getSlackThread, slackAPI, SlackConversationThread } from '../slack';
-import { getInstallationApiClient, getIntegrationInstallationForTeam } from '../utils';
+import {
+    getInstallationApiClient,
+    getIntegrationInstallationForTeam,
+    isDocsAgentsConversationsEnabled,
+} from '../utils';
 import { IngestSlackConversationActionParams } from './types';
 import { Logger } from '@gitbook/runtime';
 
@@ -21,6 +25,12 @@ export async function ingestSlackConversation(params: IngestSlackConversationAct
     const accessToken = (installation.configuration as SlackInstallationConfiguration)
         .oauth_credentials?.access_token;
 
+    const isDocsAgentsEnabled = await isDocsAgentsConversationsEnabled({
+        organizationId: installation.target.organization,
+        context,
+    });
+    const appOrgURL = installation.urls.app.replace('/integrations/slack', '');
+
     await Promise.all([
         slackAPI(
             context,
@@ -29,7 +39,9 @@ export async function ingestSlackConversation(params: IngestSlackConversationAct
                 path: 'chat.postMessage',
                 payload: {
                     channel: channelId,
-                    text: `🚀 Sharing this conversation with Docs Agent to improve your docs...`,
+                    text: isDocsAgentsEnabled
+                        ? `🚀 Sharing this conversation with Docs Agent to improve your docs...`
+                        : `✨ Docs Agent is currently in private alpha.\n\nRequest early access for your organization: ${appOrgURL}/agents`,
                     thread_ts: threadId,
                 },
             },
@@ -37,16 +49,18 @@ export async function ingestSlackConversation(params: IngestSlackConversationAct
                 accessToken,
             },
         ),
-        handleIngestSlackConversationAction(
-            {
-                channelId,
-                threadId,
-                installation,
-                accessToken,
-                conversationToIngest,
-            },
-            context,
-        ),
+        isDocsAgentsEnabled
+            ? handleIngestSlackConversationAction(
+                  {
+                      channelId,
+                      threadId,
+                      installation,
+                      accessToken,
+                      conversationToIngest,
+                  },
+                  context,
+              )
+            : Promise.resolve(),
     ]);
 }
 
--- a/integrations/slack/src/utils.ts
+++ b/integrations/slack/src/utils.ts
@@ -148,3 +148,36 @@ export function isAllowedToRespond(eventPayload: any) {
 
     return !isFromBot && !isExternalChannel;
 }
+
+/**
+ * Check if Docs agents conversation is enabled for the organization.
+ *
+ * TODO: Remove when Docs agents reached general availability.
+ */
+export async function isDocsAgentsConversationsEnabled(params: {
+    organizationId: string;
+    context: SlackRuntimeContext;
+}): Promise<boolean> {
+    const { organizationId, context } = params;
+    try {
+        const response = await fetch(
+            `https://front.reflag.com/features/enabled?context.company.id=${organizationId}&key=DOCS_AGENTS_CONVERSATIONS`,
+            {
+                method: 'GET',
+                headers: {
+                    Authorization: `Bearer ${context.environment.secrets.REFLAG_SECRET_KEY}`,
+                    'Content-Type': 'application/json',
+                },
+            },
+        );
+
+        const json = (await response.json()) as {
+            features: { DOCS_AGENTS_CONVERSATIONS: { isEnabled: boolean } };
+        };
+        const flag = json.features.DOCS_AGENTS_CONVERSATIONS;
+
+        return flag.isEnabled;
+    } catch (e) {
+        return false;
+    }
+}
--- a/integrations/slack/gitbook-manifest.yaml
+++ b/integrations/slack/gitbook-manifest.yaml
@@ -79,6 +79,7 @@ secrets:
     CLIENT_SECRET: ${{ env.SLACK_CLIENT_SECRET }}
     SIGNING_SECRET: ${{ env.SLACK_SIGNING_SECRET }}
     OPENAI_API_KEY: ${{ env.SLACK_OPENAI_API_KEY }}
+    REFLAG_SECRET_KEY: ${{ env.REFLAG_SECRET_KEY }}
 envs:
     dev-steeve:
         organization: idE5kUnGGjoPGcbu3FZJ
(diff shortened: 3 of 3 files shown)


integrations/slack/src/actions/ingestConversation.ts
```diff
@@ -0,0 +1,237 @@
...
+    try {
+        await client.orgs.ingestConversation(organizationId, conversation);
+        await slackAPI(
+            context,
+            {
+                method: 'POST',
+                path: 'chat.postMessage',
+                payload: {
+                    channel: channelId,
+                    text: `🤖 Got it! Docs Agent is on it. We'll analyze this and suggest changes if needed.`,
```
SamyPesse, line 104: Note for latr, it'd be interesting for docs agent to notify back when CR has been created in relation to this thread

integrations/slack/src/actions/ingestConversation.ts
```diff
@@ -0,0 +1,237 @@
...
+                messageTs: slackThread.messageTs,
+            },
+            createdAt: new Date(slackThread.createdAt).toISOString(),
+        },
+        parts: slackThread.messages
+            .map((message) =>
+                message.text
+                    ? ({
+                          type: 'message',
+                          role: 'user',
```
SamyPesse, line 152: Is there a way to easily distinguish here if the thread includes team members (part of the workspace) and users that are not part of the workspace?

`user` should usually be for someone who is not part of the workspace (shared slack channel with customer).

`agent` should be for an agent of the company (someone who is part of the workspace).

SamyPesse: If you have the `user` who posted the message, you can use `is_stranger` (https://docs.slack.dev/reference/objects/user-object)

integrations/slack/src/actions/ingestConversation.ts
```diff
@@ -0,0 +1,237 @@
...
+                          }),
+                    thread_ts: threadId,
+                    unfurl_links: isIngestedFromLink,
+                },
+            },
+            {
+                accessToken,
+            },
+        ),
+        queueIngestSlackConversationTask({
```
SamyPesse: Why do we need to use a queue?

I understood that for AskAI, it's because it can take a long time to respond; but ingestion of conversation should be fast as it'l be queued on the GitBook side.

integrations/slack/src/actions/types.ts
```diff
@@ -0,0 +1,74 @@
...
+    context: SlackRuntimeContext;
+
+    /* needed for postEphemeral */
+    userId?: string;
+
+    /* Get reply in thread */
+    threadId?: string;
+}
+
+interface IngestSlackConversationWithConversation extends ActionBaseParams {
```
SamyPesse: I'm slightly confused between `IngestSlackConversationWithConversation` and `IngestSlackConversationWithText`; and this PR generally lacks comment on the flow as it's quite complex.

I'd be great if you can describe a bit better the different cases in a comment :)

integrations/slack/src/handlers/actions.ts
```diff
@@ -32,10 +37,30 @@ export const slackActionsHandler = async (request: Request, context: SlackRuntim
...
+                threadId: message_ts,
+                userId: user.id,
+                context,
+                conversationToIngest: {
+                    channelId: channel.id,
+                    messageTs: message_ts,
+                },
+            };
+
+            context.waitUntil(ingestSlackConversation(params));
```
SamyPesse, line 59: since we do a waitUntil; I'm really doubting that we need the queue for ingestion.

Maybe not even for the ask AI?

integrations/slack/src/slack.ts
```diff
@@ -66,6 +84,114 @@ async function fetchChannelsByType(context: SlackRuntimeContext, types: string)
...
+        link: permalinkResponse.permalink,
+        createdAt,
+        messages,
+    };
+}
+
+/**
+ * Extracts channel ID and message timestamp from a Slack thread/conversation permalink.
+ */
+export function parseSlackConversationPermalink(permalink: string): {
```
SamyPesse: It looks risky; is there no way to get this from the API / event?

integrations/slack/src/utils.ts
```diff
@@ -109,8 +109,24 @@ export function stripBotName(text: string, botName: string) {
     return text.split(new RegExp(`^.*<@${botName}> `)).join('');
 }
 
-export function getActionNameAndType(actionId: string) {
-    const [actionName, actionPostType = 'ephemeral'] = actionId.split(':');
+export function getActionNameAndType(payload: any) {
```
SamyPesse, line 122: Can you comment quickly on why we need this? without expertise in the Slack API / app ecosystem; this is confusing

integrations/slack/src/actions/ingestConversation.ts
```diff
@@ -0,0 +1,237 @@
...
+    const installation = await getIntegrationInstallationForTeam(context, teamId);
+    if (!installation) {
+        throw new Error('Installation not found');
+    }
+
+    const accessToken = (installation.configuration as SlackInstallationConfiguration)
+        .oauth_credentials?.access_token;
+
+    const permalink = params.text;
+    const isIngestedFromLink = !!permalink;
```
SamyPesse: Related to my comment on `parseSlackConversationPermalink`.

I'm confused by why the `text` passed as params would be the link to index?

Don't we have the info on the message in the payload of the action triggered?

SamyPesse (commented):
Looks good; Slack is a complex beast so most of my comment are about understanding how it works

integrations/slack/src/actions/ingestConversation.ts
spastorelli: Yeah I had the same thought as well. We could have the integration expose a hook handler and the docs agent to send a request to that handler when the CR was created. It'd be a nice to have.

integrations/slack/src/actions/ingestConversation.ts
spastorelli: Unfortunately AKAIK the `conversation.replies` endpoint return just the user ID in the message:
https://docs.slack.dev/reference/objects/conversation-object

To get this information we would have to make a `user.info` request for each participant of a thread which can add to latency.

integrations/slack/src/actions/ingestConversation.ts
spastorelli: I thought it would be better as we also need to make the Slack API `conversation.replies` call to fetch the conversation. But yeah you're right even with that Slack call it wouldn't be as long as the askAI one. I'll change it.

integrations/slack/src/handlers/actions.ts
spastorelli: Yes I'll make the change for ingestion. For askAI I think we still do see this past conversation and why it was changed to using task:
https://gitbook.slack.com/archives/C01NXGWJELS/p1758615207688849

integrations/slack/src/actions/types.ts
spastorelli: Yes I'll add comments.
In short:

- `IngestSlackConversationWithConversation` is for cases when the ingestion was triggered via the menu shortcut like below which in this case you get the details of the conversation on which the shortcut was triggered and that ultimately the user want to ingest:
Screenshot 2025-11-03 at 21 01 53

- `IngestSlackConversationWithText` is for cases where the user invoked the `/gitbook ingest <permalink>` command using a link from a conversation elsewhere. So the `text` would be the permalink to the conversation they want to ingest. In this case, we would indeed get a `channel_id` and `message_ts` in the event payload but these would be related to the conversation the command was invoked from (and to which we want to reply) **but not** the conversation that the user wants to index.
This is to give the flexibility for the user to index a conversation thread from outside that conversation.

integrations/slack/src/slack.ts
spastorelli: Not that I could find. You can get a permalink of a conversation from its channel ID and message TS using the API (which I use for the method to get the conversation thread: info + messages). But I don't think they have one that returns a conversation channel ID + message TS based its permalink hence why I added this util.

integrations/slack/src/actions/ingestConversation.ts
spastorelli: Not necessarily, depends on how the ingestion is triggered. See my reply here:
https://github.com/GitbookIO/integrations/pull/1030#discussion_r2487723380

integrations/slack/src/actions/ingestConversation.ts
```diff
@@ -0,0 +1,195 @@
...
+import { getInstallationApiClient, getIntegrationInstallationForTeam } from '../utils';
+import { IngestSlackConversationActionParams } from './types';
+import { Logger } from '@gitbook/runtime';
+
+const logger = Logger('slack:actions:ingestConversation');
+
+/**
+ * Ingest the slack conversation to GitBook aiming at improving the organization docs.
+ */
+export async function ingestSlackConversation(params: IngestSlackConversationActionParams) {
```
SamyPesse, line 17: Should we hard-code a condition to disable this for now until we ship docs agent to everyone?

It can be a basic condition to disable in prod for now, or based on our org ID ?

integrations/slack/src/actions/ingestConversation.ts
SamyPesse: Makes sense 👍  it's not critical for now

integrations/slack/src/actions/types.ts
SamyPesse: Is `/gitbook ingest <permalink>` worth it? I find it hard to discover and also for users to know how to find the permalink

I think we should focus on:

- `@gitbook can you ingest this?` (natural language by talking to GitBook)
- The menu action

SamyPesse (approved):

integrations/slack/src/actions/ingestConversation.ts
spastorelli: The shortcut will only appear only if update the Slack app config to enable the Interactivity and add the handler to the specific shortcut settings (see screenshot).

So we can just set it up for now for staging app and not for prod. Unless we also want to test in prod in which case we need to publish another instance of a Slack integration to our prod organization (under a different name e.g `slack-private`) that uses our Slack staging app.

Image

integrations/slack/src/actions/types.ts
spastorelli: The idea for adding the shortcut was to allow customer to trigger the ingestion of a thread without having the integration post notification message to that same thread but another thread. So for instance if they were getting feedback from one of their customer in a channel they had with them but didn't want to have the thread they had with their customer poluted with notification from the itnegration. They could just trigger the command from another channel of their workspace (e.g support channel) and trigger the ingestion.

But I agree it may not be worth adding it at the moment. So removed it and in any case we could always re-add it back if we feel the need to.

integrations/slack/src/actions/ingestConversation.ts
spastorelli: Following our discussion this morning, made these changes to allow us and early adaptors test in prod:
https://github.com/GitbookIO/integrations/pull/1043

PR #1031, opened 2025-11-03
Use AI to detect intent of user when using Slack app mention

spastorelli:
Since we are adding ingestion of conversation of the Slack integration (in [this PR](https://github.com/GitbookIO/integrations/pull/1030)) and need to support it in app mention as well. So we need a good way to detect the intent of the user invoking the Slack bot via app mention. 

This PR aims at using OpenAI lightweight model to help with this.

### Demo

| **Requesting to ingest conversation**    | **Asking a question** |
| ----------- | ----------- |
| <video src="https://github.com/user-attachments/assets/0d9bb768-e311-486b-927b-ab0a630ce1b4" />      | <video src="https://github.com/user-attachments/assets/11497d22-0ce6-49c3-8027-8a3b053b785d" />       |

Based on branch from https://github.com/GitbookIO/integrations/pull/1030

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -1782,7 +1782,7 @@
 
     "events": ["events@3.3.0", "", {}, "sha512-mQw+2fkQbALzQ7V0MY0IqdnXNOeTtP4r0lN9z7AAawCXgqea7bDii20AYrIBrFd/Hx0M2Ocz6S111CaFkUcb0Q=="],
 
-    "eventsource-parser": ["eventsource-parser@3.0.0", "", {}, "sha512-T1C0XCUimhxVQzW4zFipdx0SficT651NnkR0ZSH3yQwh+mFMdLfgjABVi4YtMTtaL4s168593DaoaRLMqryavA=="],
+    "eventsource-parser": ["eventsource-parser@3.0.6", "", {}, "sha512-Vo1ab+QXPzZ4tCa8SwIHJFaSzy4R6SHf7BY79rFBDf0idraZWAkYrDjDj8uWaSm3S2TK+hJ7/t1CEmZ7jXw+pg=="],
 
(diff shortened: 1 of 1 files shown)


.github/workflows/production.yaml
```diff
@@ -61,6 +61,7 @@ jobs:
                   SLACK_CLIENT_ID: ${{ secrets.SLACK_PROD_CLIENT_ID }}
                   SLACK_CLIENT_SECRET: ${{ secrets.SLACK_PROD_CLIENT_SECRET }}
                   SLACK_SIGNING_SECRET: ${{ secrets.SLACK_PROD_SIGNING_SECRET }}
+                  SLACK_OPENAI_API_KEY: ${{ secrets.SLACK_OPENAI_API_KEY }}
```
SamyPesse: We should use 1password now and not GH secrets. See https://github.com/GitbookIO/integrations/blob/main/integrations/intercom-conversations/gitbook-manifest.yaml#L24

integrations/slack/.gitignore
```diff
@@ -0,0 +1 @@
+scripts/*.csv
```
SamyPesse: You probably don't need this anymore

integrations/slack/package.json
```diff
@@ -17,6 +20,7 @@
         "typecheck": "tsc --noEmit",
         "check": "gitbook check",
         "publish-integrations": "gitbook publish .",
-        "publish-integrations-staging": "gitbook publish ."
+        "publish-integrations-staging": "gitbook publish .",
+        "train-classifier": "bun run scripts/train-classifier.ts --"
```
SamyPesse: Not use anymore?

turbo.json
```diff
@@ -20,6 +20,7 @@
         "SLACK_CLIENT_ID",
         "SLACK_CLIENT_SECRET",
         "SLACK_SIGNING_SECRET",
+        "SLACK_OPENAI_API_KEY",
```
SamyPesse: See comment on how we don't need it anymore

PR #1054, opened 2025-11-18
Fix Front script injection

dugjason:
Fixes script injection issue. 
The script now appends to the `<body>` element as recommended by Front, and a duplicate `appendChild` call has been removed.

Refer to Front docs https://help.front.com/en/articles/2049#install_front_chat_on_your_website

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -136,7 +136,7 @@
     },
     "integrations/front": {
       "name": "@gitbook/integration-front",
-      "version": "1.0.0",
+      "version": "1.0.1",
       "dependencies": {
         "@gitbook/runtime": "*",
       },
@@ -592,7 +592,7 @@
     },
     "integrations/slack": {
       "name": "@gitbook/integration-slack",
-      "version": "2.5.3",
+      "version": "2.6.0",
       "dependencies": {
         "@ai-sdk/openai": "^2.0.62",
         "@gitbook/api": "*",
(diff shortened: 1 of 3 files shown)


addisonschultz:
Hi @dugjason 

Thanks for investigating this with us! And thanks for starting this PR. I’ve tested this out, but it’s still not working. 

After debugging a little more, it looks like it’s coming from the `color-scheme` set from the actual front widget. 

It’s currently injecting `normal`, which for some reason isn’t respecting the inherited color scheme set on the page. 

It looks like updating it to `light dark` (allowing both color schemes to go through) fixes the problem. Is this something you’d be able to look at with your product team? 

https://github.com/user-attachments/assets/a324d61f-c0ba-428d-865c-2fba9cf263b1

addisonschultz:
@dugjason actually - apologies, backpack is still running into the issue. I’m doing some more testing on my end too: 

image

addisonschultz:
Thanks @dugjason @Vilos92 - appreciate you pushing this over the line for us and Sam - I’ll finish it off and let you know if there’s any questions in https://github.com/GitbookIO/integrations/pull/1142

addisonschultz:
@taranvohra do you maybe have any context for @Vilos92 on how injected integrations are rendered? 

Context on what’s happening: The Front widget integration is rendering with a white background around the chat button on GitBook sites that are set to dark mode. It only seems to happen to sites that are force-set to dark mode (in customization settings), and not to sites that you can toggle between light/dark. 

@Vilos92 implemented a change to the Front SDK that allows us to pass `color-scheme` to the widget, but I’m worried it might be overwritten by the GitBook document

PR #1057, opened 2025-11-19
Add GitHub proxy support

conico974:
Introduce a proxy for GitHub API requests, enhancing flexibility and security. Update dependencies and implement linting for code quality.

### DIFF
--- a/.changeset/big-kiwis-wish.md
+++ b/.changeset/big-kiwis-wish.md
@@ -1,5 +1,0 @@
----
-'@gitbook/integration-intercom-conversations': patch
----
-
-Fix authentication flow when not requiring the permission "read_admins".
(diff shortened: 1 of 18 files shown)


integrations/github/src/api.ts
```diff
@@ -402,3 +417,51 @@ export function extractTokenCredentialsOrThrow(
...
+    });
+}
+
+export async function shouldUseProxy(context: GithubRuntimeContext): Promise<boolean> {
+    const companyId = context.environment.installation?.target.organization;
+    if (!companyId) {
+        return false;
+    }
+    try {
+        const response = await fetch(
```
emmerich: I think it'd be better if the proxy/no-proxy decision was managed and decided by us on the GitBook side, rather than reflag - probably something in the installation.


```suggestion
const shouldUseProxy = context.environment.installation?.proxied;
```

conico974: Yeah I agree, right now it works this way for gitlab as well. I wanted to do something quick to unblock this https://linear.app/gitbook-x/issue/RND-8673/static-ips-for-github-git-sync
We could revisit this later

PR #1178, opened 2026-04-16
Fix git sync export showing no commit message for computed revisions

spastorelli:
This PR aims to fix Git sync exports showing GitBook: No commit message when the revision is a `computed` revision.

More [context](https://gitbook.slack.com/archives/C01NXGWJELS/p1776336411577339?thread_ts=1776330439.987039&cid=C01NXGWJELS)


emmerich (approved):
It could be good to include a test on the GBX side that the revision is returned with the mergedFrom field when computed=false (if the test doesn't already exist)

PR #1185, opened 2026-05-13
Update API to support adding integrations inside a reusable content

Antoine-C:
…chema


emmerich (approved):
Can you make sure the merge commit has a clean commit message? i.e. not the one in the PR title.

PR #1191, opened 2026-05-26
Add support for icons in tabs in API

tgargula:


bun.lock
```diff
@@ -655,7 +644,7 @@
     },
     "integrations/va-auth0": {
       "name": "@gitbook/integration-va-auth0",
-      "version": "0.4.0",
+      "version": "0.5.0",
```
BrettJephson: Not sure we need to update the bun.lock.

BrettJephson (approved):

bun.lock
tgargula: it updated automatically when running the `bun install`. I can revert it

PR #1203, opened 2026-06-15
Update API

gregberge:


taranvohra (approved):

packages/runtime/src/context.ts
```diff
@@ -104,7 +104,10 @@ export function createContext(
             queueTask: async (input) => {
                 await integrationAPI.integrations.queueIntegrationTask(
                     environment.integration.name,
-                    input,
+                    {
+                        task: input.task as Record<string, string>,
```
taranvohra, line 108: It should be `Record<string, unknown>`

gregberge: It will not work with typing if I do this.

PR #1205, opened 2026-06-15
Autogenerate CLI

talagluck:
This makes `gitbook` a **spec-generated CLI**. The command tree is generated from the GitBook OpenAPI spec by `scripts/generate-commands.ts` — one command per public operation, named off the URL path (e.g. `gitbook organizations list`, `gitbook spaces get <id>`). It ships shell completion for bash/zsh/fish.

Previously this lived on a separate `gitbook2` binary while the generated commands were developed in isolation; it has now taken over the `gitbook` name and the standalone `gitbook2` binary is gone.

### What's in it

- **Generated API commands** — every public operation, grouped by resource, mounted at the top level (`gitbook organizations list`, `gitbook spaces get <id>`, …). `generated-commands.ts` is auto-generated — don't hand-edit it.
- **Compact output by default.** Dumping the full response for list commands was unreadable, so pretty output is now compact:
  - **Lists** print one line per item — id + name/title + a couple of identifying fields — with a footer showing the count and the pagination cursor as a ready-to-paste flag (`next page: --page <cursor>`).
  - **Single objects** show just the identifying fields, with a hint to use `--full`.
  - `--full` restores the complete dump; `--json` / `--yaml` are untouched, so scripts and agents still get the full structured response. When piped (not a TTY) it defaults to YAML.
  - Formatting logic lives in `packages/cli/src/output.ts` (single source of truth, unit-tested — see `output.test.ts`, `bun test`).
- **Browser (OAuth) login.** `gitbook login` authenticates through the browser using the GitBook OAuth server (dynamic client registration + PKCE + loopback redirect), with automatic token refresh; `gitbook logout` clears it. `gitbook auth --token` (personal API token) still works. The two can coexist: the OAuth session drives API commands, while a personal token is used — and required — for publishing integrations (which is not an OAuth-exposed operation). `gitbook whoami` shows which is in use.
- **Integration lifecycle commands.** The build/publish lifecycle is under `gitbook integration <verb>` (`new`, `dev`, `publish`, `unpublish`, `tail`, `check`) plus `gitbook openapi publish`. The historical top-level spellings (`gitbook publish`, …) keep working as **deprecated aliases** that print a warning pointing at the `integration` form.

### Build & publish

- Codegen is part of every build: `prebuild` runs `generate-commands` before `build.sh` bundles, and this fires under `bun run build`, `turbo run build`, and the `turbo`-driven `typecheck`/`test` (both `dependsOn: build`).
- A `prepack` script guarantees `npm publish` / `changeset publish` always regenerate + rebuild, even outside turbo.
- The OpenAPI spec is fetched live at build time (local `gitbook-x` checkout if present, else `https://api.gitbook.com/openapi.yaml`), so every build reflects the current spec. Because the CLI declares `@gitbook/api: "*"`, an API-only version bump doesn't auto-republish the CLI, so bumping the API also needs a `@gitbook/cli` changeset — enforced by the `Changeset guard` CI check and documented in [CONTRIBUTING](.github/CONTRIBUTING.md#releasing-packages).

### Open questions / follow-ups

- **Republish on spec changes.** Handled for now via a `@gitbook/cli` changeset alongside each API bump (enforced by CI). If we later want it fully hands-off, a scheduled "rebuild + release if the command tree changed" workflow is the next step.
- **Streaming operations.** The 5 SSE `text/event-stream` operations (`ask`, `ask/questions`, `ai/response`) are still skipped by the generator — they don't fit the one-response-per-command model and need a separate emit path (deferred to its own PR).
- **Usage stats** — not yet added.

## How to test

Requires Node ≥18 and bun.

```bash
# 1. Get the branch
git fetch origin autogenerate_cli
git checkout autogenerate_cli

# 2. Install deps (from the repo root)
bun install

# 3. Build the CLI (regenerates commands from the spec, then bundles)
cd packages/cli
bun run build          # runs `generate-commands` via prebuild, then build.sh

# 4. Authenticate
node dist/cli.js login              # browser (OAuth)
#   …or with a personal API token (needed for publishing integrations):
#   node dist/cli.js auth --token <YOUR_TOKEN>
```

Then run any command via `node dist/cli.js <command>`:

```bash
# Compact, readable output (the default when attached to a terminal)
node dist/cli.js organizations list
node dist/cli.js spaces list --organization <orgId>
node dist/cli.js organizations get <orgId>

# Escape hatches
node dist/cli.js organizations list --full     # full pretty dump
node dist/cli.js organizations list --json      # machine-readable (unchanged)
node dist/cli.js organizations list --yaml      # machine-readable (default when piped)

# Pagination — the footer prints the exact flag to paste
node dist/cli.js organizations list --page <cursor>

# Integration lifecycle
node dist/cli.js integration publish .

# Discover commands
node dist/cli.js --help
node dist/cli.js organizations --help
```

### Optional: install and run it as `gitbook`

```bash
# from packages/cli
bun link                 # links the `gitbook` bin onto your PATH
gitbook organizations list
```

### Run the formatting unit tests

```bash
# from packages/cli
bun test
```


packages/cli/src/generated-commands.ts
```diff
@@ -0,0 +1,11291 @@
+/**
```
spastorelli: I don't think this should belong to `src/` folder. If this is autogenerated by the generate-command.ts script does it need to be commited to the repo? Should we instead add it to the `.gitignore` or even better have the `build` process generate this file and produce a minized version under `dist/` which should already be git ignored.

scripts/generate-commands.ts
```diff
@@ -0,0 +1,1083 @@
...
+            );
+        }
+    }
+
+    return { bash, zsh, fish: fishLines.join('\n') + '\n' };
+}
+
+// ─── Entry point ──────────────────────────────────────────────────────────────
+
+const SPEC_PATH = path.resolve(__dirname, '../packages/api/spec/openapi.yaml');
```
spastorelli: Why read the yaml version instead of importing the `.json` version directly? don't we have the `.json`version included in `packages/api/spec/`?

packages/cli/src/remote.ts
```diff
@@ -1,15 +1,31 @@
...
- * Get an authenticated API client.
+ * Build the User-Agent sent to the API.
+ *
+ * `command` is the dotted command path the user invoked (e.g.
+ * "spaces.change-requests.content.update"). Including it lets requests be
+ * attributed to a specific command in the API logs / the GCP per-endpoint
+ * dashboard, without per-request plumbing. Omitted → just the surface token.
  */
-export async function getAPIClient(requireAuth: boolean = true): Promise<GitBookAPI> {
+function buildUserAgent(command?: string): string {
```
spastorelli: I don't think the UA should be based on the command. `GitBook-CLI/<version>` should be sufficient.

packages/cli/src/generated-commands.ts
```diff
@@ -0,0 +1,11291 @@
...
+        .action(async (userId, options) => {
+            const api = await getAPIClient(true, 'users.get');
+            const path = `/users/${userId}`;
+            try {
+                const response = await api.request({
+                    path,
+                    method: 'GET',
+                    secure: true,
+                });
+                const text = await response.text();
```
spastorelli: Why do we handle the response as text do then parse the JSON. I believe the `request` underlying HttpClient from Api/GitBookAPI class should handle it via the `format` param.

packages/cli/src/generated-commands.ts
```diff
@@ -0,0 +1,11291 @@
...
+        .option('--yaml', 'Output as YAML (machine-readable)')
+        .option('--pretty', 'Output in human-readable form (default when attached to a terminal)')
+        .option('--full', 'Show all fields (disable compact summaries in pretty mode)')
+        .action(async (spaceId, options) => {
+            const api = await getAPIClient(true, 'spaces.get');
+            const path = `/spaces/${spaceId}`;
+            const query: Record<string, string> = {};
+            if (options["shareKey"] !== undefined) query['shareKey'] = String(options["shareKey"]);
+            try {
+                const response = await api.request({
```
spastorelli: I think our API client classes (Api) is generated based on the resource + operation Id

Image

can't we reuse this and call the associated class instance method named using the operationId get from the spec instead of calling the underlying `.request()`? It also properly pass the `format` param that I mention in a another comment.

packages/cli/src/generated-commands.ts
spastorelli, line 1: More generally and not saying we should do this. But have you explored using SDKs/Libs that allows generating stubs based on an API spec? Weren't there suitable for our case?

spastorelli (commented):
Looks good 👌, have a few questions & comments.

packages/cli/src/generated-commands.ts
talagluck: I decided not to do this after talking to Samy, and also because part of this is autogenerated, but part (the integrations CLI) isn't.

packages/cli/src/generated-commands.ts
talagluck: Agreed  - I updated this so it's no longer tracked (git rm --cached) and is now gitignored. prebuild regenerates it from the spec before every build, so it's derived-only. I kept it emitted into src/ (rather than dist/) so the file the bundler consumes stays alongside the other CLI source and output.ts can import from it, but it's no longer committed.

packages/cli/src/generated-commands.ts
talagluck: Done - reads packages/api/spec/openapi.json via JSON.parse now, and dropped the js-yaml import.

packages/cli/src/generated-commands.ts
talagluck: Done. The one thing is that I'm missing the five endpoints that stream - I'll add these in a follow-on PR.

packages/cli/src/remote.ts
talagluck: Done, it's just `GitBook-CLI/<version>` now. Dropped the command param from getAPIClient and the per-command threading in the generator. I'd originally added it for per-endpoint attribution in the GCP dashboard, but you're right that the UA is the wrong channel.

scripts/generate-commands.ts
talagluck: Great call - updated.

packages/cli/src/oauth.ts
```diff
@@ -0,0 +1,421 @@
...
+
+/**
+ * Resolve the OAuth authorization server endpoint for a given API endpoint.
+ *
+ * The OAuth server lives on a sibling host of the API (e.g. `api.gitbook.com` ->
+ * `oauth.gitbook.com`, `api-tal.gitbook.space` -> `oauth-tal.gitbook.space`). It can be
+ * overridden explicitly with `GITBOOK_OAUTH_ENDPOINT` for setups that don't follow that
+ * convention.
+ */
+export function resolveOAuthEndpoint(apiEndpoint: string): string {
```
spastorelli, line 52: nit: The oauth server endpoint should be included in the OpenAPI spec if you want to use this for discovery. It's located under the `oauth` securityScheme with all the supported scopes:
https://api.gitbook.com/openapi.json

Image

.changeset/cli-browser-oauth-login.md
```diff
@@ -0,0 +1,5 @@
+---
+'@gitbook/cli': minor
+---
+
+Add `gitbook login` / `gitbook logout` to authenticate through the browser using OAuth, with automatic token refresh. A browser session and a personal API token can be configured at the same time: the OAuth session is used for API commands, while a personal token (via `gitbook auth --token`) is still used — and required — for publishing integrations.
```
spastorelli, line 5: >  while a personal token (via `gitbook auth --token`) is still used — and required — for publishing integrations.

For this we need to enable the integrations endpoint for OAuth. They are not at the moment. We can leave it documented like this for now until we implement it. But something to keep in mind that OAuth should be able to cover even integration endpoints.

Going to create a ticket to remind myself to do it.

spastorelli: Here it is: https://linear.app/gitbook-x/issue/RND-12041/add-oauth-support-for-integrations-api-endpoints

.github/workflows/changeset-guard.yaml
```diff
@@ -0,0 +1,64 @@
...
+            - name: Checkout
+              uses: actions/checkout@v4
+              with:
+                  # Full history so we can diff against the PR base.
+                  fetch-depth: 0
+            - name: Require a @gitbook/cli changeset when @gitbook/api is bumped
+              env:
+                  BASE_SHA: ${{ github.event.pull_request.base.sha }}
+              run: |
+                  set -euo pipefail
```
spastorelli, line 20: nit: Should we have this shell script moved to its own `.sh` script under `scripts/`?

packages/cli/src/oauth.ts
```diff
@@ -0,0 +1,421 @@
...
+    const response = await fetch(registrationEndpoint, {
+        method: 'POST',
+        headers: {
+            'content-type': 'application/json',
+            accept: 'application/json',
+            'user-agent': USER_AGENT,
+        },
+        body: JSON.stringify({
+            client_name: CLIENT_NAME,
+            client_uri: CLIENT_URI,
```
spastorelli, line 124: It would be nice to provide a logo for the CLI client registration. You can do this by providing a `logo_uri` prop pointing to the URL of the hosted logo image.

spastorelli (approved):

packages/cli/src/oauth.ts
talagluck: Just checking I read this right - you're pointing at the whole oauth block (URLs + scopes) as the thing to read from rather than hardcoding? Two different answers depending on which part:

Endpoint: I'm deriving it on purpose. The spec we bundle is always prod-only, but the CLI can be pointed at staging/dev too — deriving from the API host (plus the `GITBOOK_OAUTH_ENDPOINT` override and .well-known fallback) resolves the right server per environment, where a prod URL from the spec wouldn't.

Scopes: fair point — those are env-independent, so reading them from the `securityScheme` makes sense. Happy to do that in a follow-up.

spastorelli: Yes, the goal was to derive the OAuth server URL from the API spec. I know you're already doing endpoint discovery afterward via `/.well-known/oauth-authorization-server` (which includes the scopes too), and that's the correct approach.
That said, thinking about it more, there's no real benefit beyond just computing the server URL. It was a nitpick anyway, so it's perfectly fine as is.

talagluck: I created https://linear.app/gitbook-x/issue/RND-12043/cli-source-oauth-scopes-from-the-openapi-specs-oauth-securityscheme

PR #1219, opened 2026-06-24
Fix/plausible init

addisonschultz:


.changeset/legal-breads-cover.md
```diff
@@ -0,0 +1,5 @@
+---
+'@gitbook/integration-plausible': minor
```
jpreynat: I think that you need to remove this one as it's gonna be a major release anyway

jpreynat: Actually isn't it a patch on this one since you've merged the other PR?

jpreynat (approved):

PR #1224, opened 2026-06-29
Add region selection to Intercom integration (RND-1354)

rodrcastro:
## What

Adds a **Region** configuration option (US / EU / Australia) to the Intercom integration, mirroring how the PostHog integration lets users pick their instance address.

- **`gitbook-manifest.yaml`**: new optional `region` enum property (default `US`); CSP `form-action` extended to allow `api-iam.eu.intercom.io` and `api-iam.au.intercom.io`.
- **`src/index.ts`**: reads the configured region and injects the matching Intercom API base into the script via a new `<API_BASE>` placeholder.
- **`src/script.raw.js`**: sets `api_base` on `window.intercomSettings` — Intercom's documented way to point the widget at a non-US region.
- Added a `minor` changeset.

## Why

Per RND-1354: Intercom has [regional data hosting](https://www.intercom.com/blog/regional-data-hosting/) in the EU (and Australia), but our integration used the standard US-only web snippet without overriding the API base. EU/AU-hosted customers couldn't connect their widget. Intercom's web client supports overriding `api_base`, which is what this change does.

## Notes for reviewers

- **Backward compatible**: `region` is optional and defaults to `US`, so existing installations behave exactly as before.
- **Three regions** included (US/EU/Australia) — the issue focused on EU, but AU uses the same override mechanism.
- `connect-src` already allowed `*.intercom.io`, so the regional API hosts are covered there; only `form-action` needed the explicit additions.

## Verification

- `gitbook check` passes.
- `tsc --noEmit` introduces no new errors (6 pre-existing `packages/runtime` errors confirmed identical on clean `main`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/integrations/intercom/src/script.raw.js
+++ b/integrations/intercom/src/script.raw.js
@@ -1,8 +1,10 @@
 (function () {
     const APP_ID = '<TO_REPLACE>';
+    const API_BASE = '<API_BASE>';
 
     window.intercomSettings = {
         app_id: APP_ID,
+        api_base: API_BASE,
     };
 
     var w = window;
--- a/integrations/intercom/src/script.raw.js
+++ b/integrations/intercom/src/script.raw.js
@@ -1,6 +1,27 @@
 (function () {
     const APP_ID = '<TO_REPLACE>';
     const API_BASE = '<API_BASE>';
+    const GRANTED_COOKIE = '__gitbook_cookie_granted';
+
+    function getCookie(cname) {
+        const name = `${cname}=`;
+        const decodedCookie = decodeURIComponent(document.cookie);
+        const ca = decodedCookie.split(';');
+        for (let i = 0; i < ca.length; i++) {
+            let c = ca[i];
+            while (c.charAt(0) === ' ') {
+                c = c.substring(1);
+            }
+            if (c.indexOf(name) === 0) {
+                return c.substring(name.length, c.length);
+            }
+        }
+        return '';
+    }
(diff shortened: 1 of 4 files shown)


addisonschultz (approved):

integrations/intercom/src/script.raw.js
```diff
@@ -1,8 +1,31 @@
 (function () {
     const APP_ID = '<TO_REPLACE>';
+    const API_BASE = '<API_BASE>';
+    const GRANTED_COOKIE = '__gitbook_cookie_granted';
```
addisonschultz, line 4: @rodrcastro one thing - let’s double check if this integration uses cookies (I think it does), but this added code means the chat window will not load/appear if a user does **not** accept cookies - that might be unintended. Before we were always showing the chat window regardless of user accept/decline

rodrcastro: @addisonschultz Yeah, I didn't add this at first because we didn't have it there before, but I read Intercom's help center and it seems that the widget uses cookies, so the consent would be required. This is the article I took as a basis for the decision: https://www.intercom.com/help/en/articles/2361922-intercom-messenger-cookies

addisonschultz: Okay, the article makes sense and looks good then - feel free to merge this, just be aware if users in support are asking why it’s not showing up in certain cases

PR #1233, opened 2026-07-07
Fix OneTrust consent not syncing when visitor accepts cookies (RND-11843)

nolannbiron:
Follow-up to RND-10728. Customers reported Segment still not firing after clicking "Accept All" in OneTrust (notably in Chrome), even with the consent-gating fix in place.

The root cause is not a browser difference — it's the OneTrust → GitBook consent bridge leaving `__gitbook_cookie_granted` stuck, so every consent-gated integration (Segment, Google Analytics, Intercom, …) stayed silent. Two reinforcing defects in `script.raw.js`:

- **Premature rejection on load.** The `OptanonWrapper` re-sync ran as soon as `OnetrustActiveGroups` was defined. On a first visit that global holds the essential-only default (`,C0001,`) *before the visitor answers*, so the bridge called `onReject()` and recorded a spurious "no". Now guarded by `OneTrust.IsAlertBoxClosed()`, matching the existing pattern in `applyGPCBanner`.
- **Stale gate swallows the real click.** The `OTConsentApplied` listener was gated on the once-captured `hasRecordedConsent` snapshot, so an explicit Accept/Reject was ignored whenever a decision (including the spurious one above) was already recorded — leaving consent permanently stuck until cookies were cleared. A live user choice now always re-syncs.

Verified by exercising the actual `script.raw.js` against the key scenarios (first-time accept/reject, stale-cookie accept, returning-visitor re-sync, no-interaction): the pre-fix script fails the stuck-consent case; the fix passes all.

Note: with the bridge now correct, a Brave visitor (GPC on by default → OneTrust reject path) will not fire Segment unless they explicitly accept — this is correct consent behavior.

### DIFF
--- a/integrations/onetrust/src/script.raw.js
+++ b/integrations/onetrust/src/script.raw.js
@@ -153,7 +153,7 @@
                     if (
                         !hasRecordedConsent &&
                         w.OnetrustActiveGroups !== undefined &&
-                        typeof w.OneTrust !== 'undefined' &&
+                        w.OneTrust !== undefined &&
                         typeof w.OneTrust.IsAlertBoxClosed === 'function' &&
                         w.OneTrust.IsAlertBoxClosed()
                     ) {
--- a/integrations/onetrust/src/script.raw.js
+++ b/integrations/onetrust/src/script.raw.js
@@ -147,9 +147,16 @@
                 previousOptanonWrapper && previousOptanonWrapper();
                 try {
                     w.dispatchEvent(new Event('onetrust-banner-loaded'));
-                    // Emit initial consent for returning visitors (OnetrustActiveGroups
-                    // is populated from stored consent; OTConsentApplied won't fire)
-                    if (!hasRecordedConsent && w.OnetrustActiveGroups !== undefined) {
+                    // Re-sync stored consent for returning visitors, but only once the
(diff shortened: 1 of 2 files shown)


integrations/onetrust/src/script.raw.js
```diff
@@ -147,9 +147,16 @@
...
-                    if (!hasRecordedConsent && w.OnetrustActiveGroups !== undefined) {
+                    // Re-sync stored consent for returning visitors, but only once the
+                    // banner has actually been answered. Before that, OnetrustActiveGroups
+                    // holds the essential-only default and emitConsent would wrongly reject.
+                    if (
+                        !hasRecordedConsent &&
+                        w.OnetrustActiveGroups !== undefined &&
+                        typeof w.OneTrust !== 'undefined' &&
+                        typeof w.OneTrust.IsAlertBoxClosed === 'function' &&
+                        w.OneTrust.IsAlertBoxClosed()
```
gregberge, line 158:
```suggestion
                        w.OnetrustActiveGroups !== undefined &&
                        w.OneTrust !== undefined &&
                        typeof w.OneTrust.IsAlertBoxClosed === 'function' &&
                        w.OneTrust.IsAlertBoxClosed()
```

gregberge (approved):

PR #1244, opened 2026-07-16
CI: extract changeset guard logic into scripts/changeset-guard.sh

talagluck:
## What

Extracts the `changeset-guard` bash — added inline in #1205 — out of the workflow's `run:` block into a standalone `scripts/changeset-guard.sh`, invoked via `run: ./scripts/changeset-guard.sh`.

## Why

The guard logic ("require a `@gitbook/cli` changeset when `@gitbook/api` is bumped") was ~45 lines of bash embedded in YAML. As a script it can be linted, tested, and run locally, and the workflow file becomes trivial to read.

No behavior change - this is just a refactor. Script is committed with mode `100755`.


addisonschultz (approved):
Not tested but looks good 👍

spastorelli (approved):

PR #1247, opened 2026-07-16
CLI: generator internals — path-param flags, streaming endpoints, OAuth scope fallback (RND-12046, RND-12044, RND-12043)

talagluck:
Bucket 1 of the `mcp-cli` follow-ups from #1205 — the generator-internals cluster. **Stacked on `cli-help-discoverability` (#1245, bucket 2); retarget to `main` once that merges.**

### RND-12046 — path params as positionals *and* `--flags`
Generated commands now register each OpenAPI path parameter as an optional positional **and** a `--<name>` flag; the positional wins when both are given. Missing params produce a clear error showing both invocation forms. Applied uniformly via a new `emitCommandPreamble` shared by the simple and streaming emitters.

```
gb organizations sites list <orgId>
gb organizations sites list --organizationId <orgId>
gb organizations sites list --organizationId=<orgId>
```

### RND-12044 — the 5 SSE streaming endpoints
Streaming ops are no longer skipped by the generator — they're tagged (`Route.isStreaming`) and emitted via a dedicated path that iterates the `EventIterator` through a new `StreamRenderer` in `output.ts`:
- `--json` → NDJSON (one object per line); `--yaml` → a `---`-delimited YAML document stream
- pretty → recommended questions live, answer text incrementally, agent-response events as status lines, sources as end-of-stream citations
- SIGINT flushes and exits 130; a mid-flight error surfaces after the partial output

Commands (mechanical naming, each mirrors its URL and sits beside its buffered sibling): `organizations ask stream`, `organizations ask questions stream`, `organizations sites ask stream`, `organizations sites ask questions stream`, `organizations sites ai response stream`.

### RND-12043 — OAuth scopes from the spec (as a fallback)
The premise was stale: `gitbook login` already requests scopes from the OAuth server's `.well-known` discovery metadata (environment-correct), not a hardcoded list. This adds the spec's `oauth` securityScheme scopes as a generated fallback (`generated-oauth-scopes.ts`), used **only** when discovery omits `scopes_supported`, so login never silently requests zero scopes. The OAuth endpoint stays runtime-derived (the bundled spec is prod-only).

### Idle request timeout (follow-on from testing)
All generated commands (buffered, merged, streaming) now abort a request that produces no response — for streams, no next event — within `CLI_TIMEOUT_MS` (default 60s, `GITBOOK_CLI_TIMEOUT_MS` to override, `0` disables). The stream timer resets per event so a slow-but-live response isn't cut off. This came out of testing: the org-level AI `ask` endpoints were observed hanging **server-side** (raw `curl -i -N` returns no HTTP status), so the CLI now fails loudly instead of hanging.

## Testing
- `bun run generate-commands`, `tsc --noEmit`, `./build.sh`, `bun test` (49 pass) all clean.
- Exercised end-to-end: all three path-param forms + precedence + missing-param error; streaming command wiring (reaches the correct URL, errors promptly on a bad token); timeout fires on both buffered and streaming, and the normal error path still reports real API errors.
- Not verified against a live successful AI stream — the ask endpoints were unresponsive server-side during testing (see above). The renderer is covered by unit tests across all event shapes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


packages/cli/src/output.ts
```diff
@@ -321,3 +365,181 @@ export function printResult(data: unknown, options: OutputOptions): void {
...
+}
+
+// Human-readable streaming. The three event shapes these endpoints emit are each
+// rendered on the fly:
+//   - recommended-question events (`{ question }`)   → one bullet per question
+//   - answer snapshots (`{ type: 'answer', answer }`) → the answer text, printed
+//     incrementally as the snapshot grows, then its sources as citations at end
+//   - agent-response events (`{ type: 'response_*' }`) → a concise status line
+// Unknown shapes fall back to a compact block so nothing is silently dropped.
+function createPrettyStreamRenderer(): StreamRenderer {
```
spastorelli: Shouldn't the stream renderer be agnostic of the schema of the events it outputs? My concern is that if the schema of the events would change (hopefully not but we never know 😓 ) or if we were other events with different shape then this logic should change to accomodate the new schemas. Isn't there a way to make this renderer more generic?

Maybe making the core logic of the renderer generic and accept a `StreamEventRenderer` which can be passed specifically at call site (where the API endpoint is called and response parsed) for each type of events.

spastorelli (approved):

packages/cli/src/output.ts
talagluck: Updated!

PR #1258, opened 2026-08-03
Bump API

BrettJephson:
Bump API for new InlineButton action


conico974 (approved):
You'll need to bump the cli as well

PR #1275, opened 2026-08-19
Support Amplitude projects with EU data residency

ManuelGerding:
## Problem

The integration loads the Amplitude autocapture bundle from `https://cdn.amplitude.com/script/<API_KEY>.js` unconditionally. For Amplitude projects with **EU data residency** that CDN responds with `401 "Invalid Key."` — the loader fails silently, `amplitude.init()` is never called, and the integration tracks nothing, with no visible error to the site owner.

Repro: `curl https://cdn.amplitude.com/script/<EU_KEY>.js` → 401 "Invalid Key.", while the same key on `cdn.eu.amplitude.com` → 200.

## Fix

Adds a **Server Region** configuration option (same pattern as the Mixpanel integration): selecting EU loads the bundle from `cdn.eu.amplitude.com` and sets `serverZone: 'EU'`, so event ingestion, remote config, and session replay all use the EU endpoints. A custom Server URL (e.g. a first-party proxy) is still honored in both regions; a Server URL matching the selected region's default is omitted as redundant. Existing US installations are unaffected.

- `gitbook-manifest.yaml`: new `server_region` enum (default `Standard (US)`), EU hosts added to the CSP
- `amplitudeScript.raw.js`: CDN host is a template parameter instead of a hardcoded URL
- `index.ts`: region → CDN host mapping and `serverZone` handling
- `tests/script.test.ts`: covers US default (unchanged output), EU CDN + serverZone, region-default server URL suppression, custom server URL passthrough

## Testing

- `bun test`: 4 new tests pass; typecheck/lint/format clean
- Verified end-to-end against a live GitBook site (docs.steadybit.com) with an EU Amplitude project by injecting the exact script the updated handler generates: the EU bundle loads, session replay posts to `api-sr.eu.amplitude.com`, and events are ingested by Amplitude EU.

### DIFF
--- a/.changeset/witty-regions-travel.md
+++ b/.changeset/witty-regions-travel.md
@@ -1,0 +1,5 @@
+---
+"@gitbook/integration-amplitude": minor
+---
+
+Support Amplitude projects with EU data residency: a new "Server Region" option loads the SDK from the EU CDN and routes events to the EU endpoint. Previously the injected loader always used the US CDN, which rejects EU API keys with 401 "Invalid Key.", so the integration silently never initialized for EU projects.
--- a/integrations/amplitude/gitbook-manifest.yaml
+++ b/integrations/amplitude/gitbook-manifest.yaml
@@ -15,10 +15,13 @@ scopes:
     - site:script:cookies
 contentSecurityPolicy:
     script-src: |
-        https://cdn.amplitude.com;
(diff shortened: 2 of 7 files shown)


addisonschultz (approved):

addisonschultz:
Thanks for adding this contribution @ManuelGerding !

PR #1280, opened 2026-08-25
Update the CLI to use OAuth for integrations

spastorelli:
This PR aims to update the CLI to use OAuth for integrations, so `gitbook login` also covers the integrations commands now that the API exposes integration endpoints to OAuth tokens (in https://github.com/GitbookIO/gitbook-x/pull/25029).


talagluck (approved):
LGTM!

PR #1288, opened 2026-09-03
Bump API

nolannbiron:


.changeset/better-moles-grow.md
```diff
@@ -0,0 +1,5 @@
+---
+'@gitbook/api': minor
```
conico974, line 2: You need to bump the cli as well

conico974 (approved):

spastorelli (approved):

.changeset/better-moles-grow.md
nolannbiron: I always forget 😅

PR #1290, opened 2026-09-04
Add report feedback command to GitBook CLI

spastorelli:
This PR aims to add `feedback` command to GitBook CLI to allow agents report friction with the CLI, submitting to the new agent feedback API endpoint.

### DIFF
--- a/packages/cli/src/feedback.ts
+++ b/packages/cli/src/feedback.ts
@@ -1,0 +1,215 @@
+import { GitBookAPI } from '@gitbook/api';
+
+import { config, getAuthConfig } from './config';
+import { getEnvironment } from './environments';
+import { USER_AGENT } from './remote';
+
+/**
+ * Published docs site collecting the feedback, per API endpoint.
+ *
+ * The site is chosen from the endpoint rather than a flag so a CLI pointed elsewhere can
+ * never report into the production docs site. An endpoint that is not listed collects
+ * nothing.
+ */
+const DOCS_SITE_URL_BY_API_HOST: Record<string, string> = {
+    'api.gitbook.com': 'https://gitbook.com/docs',
+};
+
+/**
+ * Refresh a cached token this long before it expires, so a submission started just under the
+ * wire does not race the expiry.
+ */
+const EXPIRY_MARGIN_MS = 5 * 60 * 1000;
+
+export interface FeedbackInput {
+    category: string;
+    summary: string;
+    command?: string;
+    flags?: string[];
+    severity?: string;
+    detail?: string;
+    goal?: string;
+    workaround?: string;
+    model?: string;
+    onDemand?: boolean;
+}
+
+/**
+ * The recorded submission, as returned by the API.
+ */
+export interface FeedbackResult {
+    id: string;
+    result: string;
+}
+
+/**
+ * Report feedback with a CLI command to GitBook's docs site.
+ */
+export async function reportFeedback(input: FeedbackInput): Promise<FeedbackResult> {
+    const { endpoint } = getAuthConfig();
+    const siteUrl = getDocsSiteURL(endpoint);
+    if (!siteUrl) {
+        throw new Error(
+            `Feedback is not collected for the API endpoint "${endpoint}", so there is nowhere to send this report.`,
+        );
+    }
+
+    const { token, organization, site } = await getContentAPIToken(endpoint, siteUrl);
+
+    // Authenticated with the docs site's own content token, not the user's credentials: the
+    // person running the CLI has no permission on the site collecting the feedback.
+    const api = new GitBookAPI({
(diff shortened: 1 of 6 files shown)


packages/cli/src/feedback.ts
```diff
@@ -0,0 +1,215 @@
...
+    return typeof expiresAt === 'number' && expiresAt - EXPIRY_MARGIN_MS > now;
+}
+
+/**
+ * A content API token for the docs site, reusing the cached one until it nears expiry.
+ *
+ * The token lives about a day, so resolving is rare; it is cached because resolving on every
+ * report would be a wasted round trip, not because the call is expensive.
+ */
+async function getContentAPIToken(
```
conico974, line 174: Why are we using a content API token here ? It seems wrong to do that really, and these calls wil get improperly categorized at the API level as coming from GBO

spastorelli: Because the report agent feedback endpoint is behind auth, so it's either a PAT or a content API token. (Still looking at whether we should open the endpoint up to no auth or make the auth configurable by customers i.e auth vs no auth but until then we would need to gather data to see if that makes sense)

A PAT can't be used for the CLI as the source is public and it ships on user machines, so there's nowhere to put one (since it would be one of our PAT).

The end user's own token doesn't work either, because the auth has to be scoped to the site that collects the feedback (in the case of the CLI it should be ours), not to whoever is running the command. We report into our own docs site, and an end user has no permission on it, so their token gets a 403. A content API token for `gitbook.com/docs` is the only credential an end user can legitimately obtain for that site, which is why it's the one we use.

On the categorisation, where would that be happening, and what's the impact? The requests should carry `User-Agent: GitBook-CLI/<version>`, so they are distinguishable from GBO.

If we really don't want to use a content API token here then happy to not add the report command to the CLI for now until we figure a way to do it properly auth wise, (and only do it for the MCP) but thought it could be good to have for the CLI too.

packages/cli/src/feedback.ts
```diff
@@ -0,0 +1,215 @@
...
+/**
+ * Published docs site collecting the feedback, per API endpoint.
+ *
+ * The site is chosen from the endpoint rather than a flag so a CLI pointed elsewhere can
+ * never report into the production docs site. An endpoint that is not listed collects
+ * nothing.
+ */
+const DOCS_SITE_URL_BY_API_HOST: Record<string, string> = {
+    'api.gitbook.com': 'https://gitbook.com/docs',
+};
```
gregberge, line 16: Why are we reporting it on our docs site? I don't think it's the same use case a feedback on docs and a feedback on the CLI.

gregberge (approved):
I think it's a good idea to provide feedback through the CLI.

packages/cli/src/feedback.ts
spastorelli: The site is the repository for the feedback, not the subject of it. The endpoint is meant to let a customer collect headless feedback about their product (through their MCP server, their CLI, their API) and their docs site is just where that feedback lands and who owns it. So one site can gather both "this page is wrong", "this documentation is incoplete" and "this command was confusing", because both are feedback an agent hit while working with that product. Scoping it to the site also fits the site-centric direction we're taking.

For now we're mostly exploring what data we can usefully collect from agents and LLMs & if this a direction we want to take. And we're using our main docs site as the repository for it.

Here's the spec of the agent feedback endpoint to collect the feedback to see more details: https://github.com/GitbookIO/gitbook-x/pull/25172

gregberge: But it's feedback relative to the documentation, not the product itself. It seems we are using GitBook as a hub to collect everything. We are not Amplitude.

spastorelli: But I don't think we're trying to collect everything. As you can see in the linear project description, the goal is to turn product knowledge into something actionable.

Missing knowledge and missing capability can often have the same signal from two sides of things. An agent that couldn't do something either hit a gap in the product or a gap in what's documented about it, and usually it can't tell which. That's the sort of thing a docs platform is well positioned to catch, which is why this could fits within or next to content gaps.

Agreed we're not Amplitude and I don't think we are trying to be. But there can be an opportunity for us to help customers gather this qualitative feedback and extract actionable gaps for them.

And again this is something we are exploring right now so nothing set in stone. We want to see what can come out of it by using it with our MCP or CLI before deciding to make a feature.

gregberge: Feel free to ignore me on this @spastorelli, it's fine to log things in GitBook.

spastorelli: No it's good feedback I don't necessarily agree that we are trying to be like Amplitude with this but yeah this is something we need to keep in mind when we explore this.

